PR #37, opened 2018-06-11
[WIP] Prototype of jaeger tracing

guewen:
It uses the opencensus (https://opencensus.io) python library with the jaeger exporter.

To test:
* Start a all-in-one Docker image for the jaeger server (https://www.jaegertracing.io/docs/getting-started/#all-in-one-docker-image)
* install git in odoo/Dockerfile (for the installation of opencensus from source)
* Install the dependencies, add in odoo/requirements.txt

    futures
    git+git://github.com/census-instrumentation/opencensus-python#egg=opencensus
    thrift

* Install the module
* Configure the environment variable

      - ODOO_OPENCENSUS_TRACING=1

* Run an jaeger agent in the docker-composition it would be (add `jaeger-agent` in the `depends_on` of `odoo`):

      jaeger-agent:
        image: jaegertracing/jaeger-agent:latest
        command: /go/bin/agent-linux --collector.host-port=jaeger:14267

      jaeger:
        image: jaegertracing/all-in-one:latest
        environment:
          - COLLECTOR_ZIPKIN_HTTP_PORT=9411
        ports:
          - 16686:16686

When you start odoo, it will begin to
send tracing data to the jaeger agent (in UPD) which will in turn send
tracing data to the server.

What's not working or still to do

* Tracing of models methods: if we want to trace them, we'll have to
find a trick to have the tracing include *all* the overrides of all
the addons. It should be the first method in the MRO with super calling
the others...
* IrHttp._dispatch is probably too late in the call chain because it doesn't include the authentication, we should start the span the earliest possible
* Configuration
* Better tags

selection_305


guewen:
Had to delete my fork to reattach after the repository went to private to public, so this branch is orphan now. I will need to create a new PR when I can progress on it.

PR #42, opened 2018-12-11
allow more bucket names

tschanzt:
Allow first part of bucketnames to contain additional chars. So more names are possible like c2c-odoo-prod :smile:


guewen (approved):
Not sure why the name of the project is not camptocamp like the github project but required change in any case

tschanzt:
@guewen updated to set boto cfg to /dev/null otherwise travis doesn't pass

PR #48, opened 2019-03-01
12.0 attachment_s3 improvements

grindtildeath:
Replaces :
 - #46 
 - #47

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-boto==2.42.0
+boto3==1.9.102
 redis==2.10.5
 python-json-logger==0.1.5
 statsd==3.2.1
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,4 +1,5 @@
 boto==2.42.0
```
yvaucher: is `boto` still necessary ?

yvaucher (approved):
LGTM

Just a question about old boto.

TIL usage of _logger.exception (which is basically error with passing content of exc_info to be used in exception handler)

requirements.txt
grindtildeath: Good catch, I removed it.

PR #49, opened 2019-03-08
[12.0] Add base_fileurl_field

grindtildeath:
Requires :
 - [x] #48


blaggacao:
@grindtildeath Interesting module. Is it value to you to suggest a bundled js widget implementation?

grindtildeath:
@blaggacao I'm not sure what you mean by "bundled js widget" ? :no_mouth:

blaggacao:
@grindtildeath Oh sorry, I wasn't utterly clear.

By "bundled" I wanted to refer to "bundled with this module".

I was thinking of something like:

- A JS widget which let's either upload a file or set the URL - in case Odoo is not the authoritative system which manages the life cycle of that piece of binary data.

yvaucher:
@grindtildeath #48 was merged I'll rebase to retrigger travis

test_base_fileurl_field/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    'name': 'test web url fields',
```
yvaucher: To rename after the module it tests

yvaucher (changes requested):
Test are red.

And question aside of the review, how does it play with related?

Should we read the filename on the original object ? or do I need to create a second related field only for the filename?

base_fileurl_field/fields.py
```diff
@@ -0,0 +1,102 @@
...
+                    'type': 'binary',
+                    'datas': value,
+                }
+                fname = False
+                if self.filename:
+                    fname = record[self.filename]
+                    vals['datas_fname'] = fname
+                    if fname and self.storage_path:
+                        storage_key = self._build_storage_key(
+                            record[self.filename]
```
yvaucher:
```suggestion
                            fname
```

base_fileurl_field/fields.py
```diff
@@ -0,0 +1,102 @@
...
+                ).create(vals)
+
+    def write(self, records, value):
+        for record in records:
+            storage_key = False
+            if self.filename:
+                fname = record[self.filename]
+                if fname and self.storage_path:
+                    storage_key = self._build_storage_key(
+                        record[self.filename])
```
yvaucher:
```suggestion
                        fname
```

grindtildeath:
@yvaucher I added some related fields in the test/example module, and it appears to be working although the related fields is not readonly by default as is the case with other fields types.

@blaggacao The idea here was to delegate the storage to any module implementing `base_attachment_object_storage` methods (in our case S3) and not set an URL directly in the widget. Still, a better widget to handle the filename would be great, but it's not in our plans at the moment.

base_fileurl_field/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2012-2019 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Base FileURL Field",
+    "summary": "Implements of FileURL type fields",
```
TDu: s/ Implements / Implementation

base_fileurl_field/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    "category": "Technical Settings",
+    "description": """
+        This module adds a new field type FileURL to odoo.
```
TDu: Odoo

base_fileurl_field/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+        FileURL is an extension of field type Binary, with the aim to store its
+         value on any kind external storage.
```
TDu: any kind of

base_fileurl_field/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+        It's been built with the focus on Amazon S3 but could be used with
+         other storage solution as long as it extends the functionaly of
```
TDu: s/ functionaly / functionality

test_base_fileurl_field/models/res_partner.py
```diff
@@ -0,0 +1,44 @@
...
+    url_file_fname = fields.Char()
+
+    url_image = fields.FileURL(
+        storage_location='s3',
+        filename='url_image_fname',
+        storage_path='partner_image',
+    )
+    url_image_fname = fields.Char()
+
+    @api.constrains('url_file', 'url_file_fname')
```
TDu, line 26: As I guess we would never want a duplicate url for a file, should/could this check be implemented on the field class definition ?

grindtildeath: It's unfortunately not possible at field level since the filename is implemented at model level. So the constraint must be implemented at model level as is the case there...

grindtildeath:
@TDu comments attended.

test_base_fileurl_field/models/res_partner.py
TDu: Got it, thank you.

TDu (approved):
Seems to work as advertised :+1:

yvaucher (approved):

PR #52, opened 2019-04-24
Fix S3 bucket name check to allow numbers

p-tombez:
Fix regex to allow names with numbers like "init7"


sebalix (approved):

vrenaville (approved):
:+1: Thanks

i-vyshnevska (approved):

PR #54, opened 2019-05-02
[11.0] Add method to force storage of special attachments to DB

guewen:
Some attachments (e.g. image_small, image_medium) are stored in DB
instead of the object storage for faster access.

In some situations, we may have pushed all these files on the Object
Storage (migration from a filesystem to object storage) and want to
bring back these attachments from the object storage to the database.

This method is not called anywhere but can be called by RPC or scripts.

### DIFF
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -316,7 +316,7 @@ class IrAttachment(models.Model):
                 # write them back in the DB (the logic for location to write is
                 # in the 'datas' inverse computed field)
                 attachment.write({'datas': attachment.datas})
-                # as the file will potentially be dropped on the bucket,
+                # as the file will potentially be deleted from the bucket,
                 # we should commit the changes here
                 new_env.cr.commit()
                 if current % 100 == 0 or total - current == 0:
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -6,11 +6,14 @@ import base64
 import inspect
 import logging
 import os
+import time
+
 import psycopg2
 import odoo
 
 from contextlib import closing, contextmanager
 from odoo import api, exceptions, models, _
+from odoo.osv.expression import AND, normalize_domain
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -239,6 +267,65 @@ def force_storage(self):
...
+                # access to 'datas' will compute all the 'datas' fields at
+                # once, which means reading hundreds or thousands of files at
+                # once, exhausting memory
+                attachment = model_env.browse(attachment_id)
+                # this write will read the datas from the Object Storage and
+                # write them back in the DB (the logic for location to write is
+                # in the 'datas' inverse computed field)
+                attachment.write({'datas': attachment.datas})
+                # as the file will potentially be dropped on the bucket,
+                # we should commit the changes here
```
yvaucher, line 320: I don't get the link between the bucket and the change in DB

I don't understand the meaning of  "dropped on", is it:

A. A typo to say dropped from as in Deleted from the bucket?
B. To mean pushed into the bucket?

My guess is that it is B. but I don't get why. From what I understand we want to give shelter to the files into the database and not kick them out.
So it should be closer to A, but then do we really want to delete the files from S3 I would expect to not erase them right away. in this case I just spotted a typo.

guewen: A. deleted from the bucket. This is how the inverse field works in ir_attachment, when you changes the data location, it writes in the new and deletes the old location.
Yes, there is a risk to lose the files if the commit fails. But all in all, the small/medium files can be rebuilt from the image so I'm not sure we want to make it too complex. (Deleting the files after would mean we have to keep a log of files to garbage collect)

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -239,6 +267,65 @@ def force_storage(self):
...
+                # if we browse attachments outside of the loop, the first
+                # access to 'datas' will compute all the 'datas' fields at
+                # once, which means reading hundreds or thousands of files at
+                # once, exhausting memory
+                attachment = model_env.browse(attachment_id)
+                # this write will read the datas from the Object Storage and
+                # write them back in the DB (the logic for location to write is
+                # in the 'datas' inverse computed field)
+                attachment.write({'datas': attachment.datas})
+                # as the file will potentially be dropped on the bucket,
```
guewen:
```suggestion
                # as the file will potentially be deleted from the bucket,
```

guewen:
Merged in 12.0

PR #62, opened 2019-05-10
7.0 backport legacy

p-tombez:
Backport modules for legacy support

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,134 @@
+[![Build Status](https://travis-ci.com/camptocamp/odoo-cloud-platform.svg?token=Lpp9PcS5on9AGbp76WKB&branch=8.0)](https://travis-ci.com/camptocamp/odoo-cloud-platform)
+
+# Odoo Cloud Addons
+
+Camptocamp odoo addons used on our Cloud Platform.
+
+## Introduction
+
+On the platform we want to achieve having:
+
+* No data stored on the local filesystem so we can move an instance
+  between hosts and even have several running front-ends
+* Logs sent to ElasticSearch-Kibana structured as JSON for better searching
+
+For the storage, we store all the attachments on an object storage such as S3 or
+Swift, and we store the werkzeug sessions on Redis.
+
+## Setup
+
+### Python dependencies
+
+Libraries that must be added in ``requirements.txt``:
+
+```
+redis==2.10.5
+python-json-logger==0.1.5
+
+# For S3 object storage (Exoscale, AWS)
+boto==2.42.0
+
+# For Swift object storage (Openstack, OVH)
+python-swiftclient==3.4.0
+python-keystoneclient==3.13.0
+```
+
+### Odoo Startup
+
+The `--load` option of Odoo must contains the following addons:
+
+* `session_redis`
+* `logging_json`
+
+Example:
+
+`--load=web,web_kanban,session_redis,logging_json`
+
+### Server Environment
+
+The addon `cloud_platform` is an addon that we use for 2 things:
+
+* validate that we setup the required environment variables depending on the running environment
+* install and configure the cloud addons
+
+For this purpose, we use the `server_environment` with the following envs:
+
+* `prod`
+* `integration`
+* `test`
+* `dev`
+
+The exact naming is important, because the `cloud_platform` addon rely on these keys to know and check the running environment.
+
+
+### Attachments in the Object Storage
+
+* prod: stored RW in the object storage
+ * `AWS_HOST`: depends of the platform
+ * `AWS_ACCESS_KEY_ID`: depends of the platform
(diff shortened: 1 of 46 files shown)


attachment_swift/README.rst
```diff
@@ -0,0 +1,51 @@
+Attachments on Swift storage
+============================
+
+This addon enable storing attachments (documents and assets) on OpenStack Object Storage (Swift)
```
asaunier:
```suggestion
This addon enables storing attachments (documents and assets) on OpenStack Object Storage (Swift).
```

base_attachment_object_storage/README.rst
```diff
@@ -0,0 +1,7 @@
+Base class for attachments on external object store
+===================================================
+
+This is a base addon that regroup common code used by addons targeting specific object store
```
asaunier:
```suggestion
This is a base addon that regroups common code used by addons targeting specific object store.
```

asaunier (approved):
Same comment than for 8.0 branch.

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -0,0 +1,298 @@
...
+                    full_path, exc_info=True
+                )
+
+
+@contextmanager
+def savepoint(cursor):
+    name = uuid.uuid1().hex
+    cursor.execute('SAVEPOINT "%s"' % name)
+    try:
+        yield
```
Tonow-c2c, line 46: we didn't need to add the expression/value to yield after one `yield` ? [https://docs.python.org/3/reference/simple_stmts.html#yield](https://docs.python.org/3/reference/simple_stmts.html#yield)

attachment_s3/models/ir_attachment.py
```diff
@@ -0,0 +1,164 @@
...
+                _logger.exception(
+                    'Error during storage of the file %s' % filename
+                )
+                raise except_orm(
+                    _('S3 Error'),
+                    _('The file could not be stored: %s') %
+                    (self._parse_s3_error(error),)
+                )
+        else:
+            _super = super(IrAttachment, self)
```
Tonow-c2c, line 137: mybe add a `TODO: refactor later`

Tonow-c2c (approved):
It's not a good idea to do like one template global and just change specific things for each version ?
Cause V7 and V8 seems pretty close.

base_attachment_object_storage/models/ir_attachment.py
p-tombez: It's not mandatory as explained here : https://docs.python.org/3/reference/expressions.html#yieldexpr
But be careful as the "official" use of `yield` is to make generators, in the following doc, you can see how it is used as a context manager (to be use in `with` statements) and a few examples of `yield` usage without value : https://docs.python.org/3/library/contextlib.html

attachment_s3/models/ir_attachment.py
p-tombez: Why ? If I do so, we should also update all other versions as it's the same everywhere

p-tombez:
> It's not a good idea to do like one template global and just change specific things for each version ?
> Cause V7 and V8 seems pretty close.

Yeah, so as v9, 10, 11 and 12. If you want to propose an improvement, go ahead but it's not the goal of this current work

base_attachment_object_storage/models/ir_attachment.py
guewen: @Tonow-c2c There is no value to yield here. 
About @p-tombez 's remark: the context manager in fact expects and uses a generator that yields only one item.

attachment_s3/models/ir_attachment.py
guewen: There is no reason to change this. It has been done  because of the line length.

guewen:
> Yeah, so as v9, 10, 11 and 12. If you want to propose an improvement, go ahead but it's not the goal of this current work

I agree, except on the "go ahead" part. Yes it is a pity when we have to apply the same fixes on every version, but in any case you would have to test them on every version and then the "small diffs for each version" can be hard to digest. I would add that there have been many tentative to have addons supporting several odoo versions / propositions that odoo support that, and no good outcome ever came out.

guewen:
@p-tombez any special part we need to look at where you needed to change code to adapt to 7.0?

README.md
```diff
@@ -0,0 +1,134 @@
+[![Build Status](https://travis-ci.com/camptocamp/odoo-cloud-platform.svg?token=Lpp9PcS5on9AGbp76WKB&branch=8.0)](https://travis-ci.com/camptocamp/odoo-cloud-platform)
```
simahawk: branch=7.0

simahawk (commented):
@p-tombez ping

PR #64, opened 2019-05-31
Fix inverse datas for swift storage

### LINKED ISSUE #63: _inverse_datas not callled
base_attachment_object_storage in v9 defines a function called _inverse_datas. The default inverse function for datas (https://github.com/odoo/odoo/blob/9.0/openerp/addons/base/ir/ir_attachment.py#L279) is _data_set so it will not be called. 

We define a datas field in attachment_s3 which uses _inverse_datas a inverse.  Can we move it? currently there is no dependency between the two afterwards attachment_s3 would have to depend on base_attachment_object_storage

tschanzt:
Fixes #63 
Checked locally by setting my env to write to a test swift container and setting all images to be written on swift. Then i ran the script which should move it to the db and it worked. It also works properly when its db storage from the beginning.

### DIFF
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -9,7 +9,7 @@ import psycopg2
 import odoo
 
 from contextlib import closing, contextmanager
-from openerp import api, exceptions, models, _
+from openerp import api, exceptions, fields, models, _
 from openerp import SUPERUSER_ID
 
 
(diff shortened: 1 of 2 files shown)


guewen (approved):
Thanks!

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -40,6 +40,19 @@ class IrAttachment(models.Model):
 
     _local_fields = ('image_small', 'image_medium', 'web_icon_data')
 
+    datas = fields.Binary(
+        compute='_compute_datas',
+        inverse='_inverse_datas',
+        string='File Content',
+        nodrop=True,
```
simahawk, line 47: curious: where this param comes from? :thinking:

simahawk (approved):
seems ok

base_attachment_object_storage/models/ir_attachment.py
tschanzt: @simahawk sorry didn't see that. It seems that it is so the column doesn't get droped when there is a function field with the same name. Which the original datas field is. also it was already in attachment_s3 :smiling_imp:

PR #71, opened 2019-07-23
FIX: use Registry intead of RegistryManager.get starting from Odoo 11.0

sebalix:


jcoux (approved):
Seems too easy, but :+1:

TDu (approved):

guewen:
We can remove the migration script, it was for 10.0.
If you need to run this migration, there is #54.

PR #73, opened 2019-08-02
monitoring_status filter: ignore HTTP version

simahawk:


simahawk:
hmm seems to have no effect :/

simahawk:
> hmm seems to have no effect :/

a stupid quote more to remove...

guewen (approved):

PR #81, opened 2019-10-07
BSRD-286: Set the addons to uninstallable

Tonow-c2c:
Related to https://jira.camptocamp.com/browse/BSRD-286

PR #82, opened 2019-10-07
Bsrd 286 mig 13.0 attachment s3

Tonow-c2c:
Related to https://jira.camptocamp.com/browse/BSRD-286

Should merge before https://github.com/camptocamp/odoo-cloud-platform/pull/81


attachment_s3/migrations/13.0.0.0.1/post-migration.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016-2018 Camptocamp SA
+# Copyright 2016-2019 Camptocamp SA
```
grindtildeath, line 1: I'm not sure if we should keep this file? :thinking: 
cc @guewen

grindtildeath (approved):
Out of the question, LGTM

asaunier (approved):

PR #97, opened 2019-10-21
Support IAM role

TelmoSenseFly:
We want boto to be able to assume role provider when credentials are not passed as parameters

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#iam-role

### DIFF
--- a/attachment_s3/models/ir_attachment.py
+++ b/attachment_s3/models/ir_attachment.py
@@ -51,10 +51,11 @@ class IrAttachment(models.Model):
         secret_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
         bucket_name = name or os.environ.get('AWS_BUCKETNAME')
 
-        params = {
-            'aws_access_key_id': access_key,
-            'aws_secret_access_key': secret_key,
-        }
+        if access_key and secret_key:
+            params = {
+                'aws_access_key_id': access_key,
+                'aws_secret_access_key': secret_key,
+            }
         if host:
             params['host'] = host
         if region_name:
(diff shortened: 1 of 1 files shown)


attachment_s3/models/ir_attachment.py
```diff
@@ -51,10 +51,11 @@ def _get_s3_bucket(self, name=None):
         secret_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
         bucket_name = name or os.environ.get('AWS_BUCKETNAME')
 
-        params = {
```
simahawk, line 54: `params` would be undefined -> broken

attachment_s3/models/ir_attachment.py
```diff
@@ -63,19 +64,6 @@ def _get_s3_bucket(self, name=None):
             params['region_name'] = region_name
         else:
             connect_s3 = boto.connect_s3
-        if not (access_key and secret_key and bucket_name):
```
simahawk, line 66: you cannot trash this so easily :)
Options:
1. override `_get_s3_bucket` in your own code. Easy and fast. No clash w/ us.
2. if you think this is something worth to support generically, you could add an extra env var like `AWS_IAM` or similar. Then if this var is present you bypass this part.

simahawk (changes requested):

attachment_s3/models/ir_attachment.py
TelmoSenseFly: Option 1, I will override ```_get_s3_bucket``` ;)

TelmoSenseFly:
@simahawk

PR #98, opened 2019-10-25
[9.0] [IMP] monitoring_status: check the db is reachable

gurneyalex:
make a real SQL query on GET /monitoring/status and return a 503
if the query failed.

### DIFF
--- a/monitoring_status/controllers/main.py
+++ b/monitoring_status/controllers/main.py
@@ -5,9 +5,12 @@
 import logging
 import json
 
+import psycopg2
+
 import werkzeug
 
 from openerp import http
+from openerp.http import request
 from openerp.addons.web.controllers.main import ensure_db
 
 
@@ -31,10 +34,24 @@ class Monitoring(http.Controller):
     @http.route('/monitoring/status', type='http', auth='none')
     def status(self):
         ensure_db()
+        http_status = 200
         # TODO: add 'sub-systems' status and infos:
         # queue job, cron, database, ...
         headers = {'Content-Type': 'application/json'}
         info = {'status': 1}
+        # check the database connection
+        try:
+            cr = request.env.cr
+            cr.execute(
+                'SELECT value '
+                'FROM ir_config_parameter '
+                'WHERE key=%s',
+                ('web.base.url',))
+            result = cr.fetchone()
+            info['web.base.url'] = result or ''
+        except psycopg2.OperationalError as exc:
+            info['database_error'] = str(exc)
+            http_status = 503
         session = http.request.session
         # We set a custom expiration of 1 second for this request, as we do a
         # lot of health checks, we don't want those anonymous sessions to be
@@ -44,4 +61,6 @@ class Monitoring(http.Controller):
         # Redis.
         if not session.uid:
             session.expiration = 1
-        return werkzeug.wrappers.Response(json.dumps(info), headers=headers)
+        return werkzeug.wrappers.Response(
+            json.dumps(info), status=http_status, headers=headers
+        )


gurneyalex:
related to BSQOQ-276

sebalix (approved):

monitoring_status/controllers/main.py
```diff
@@ -31,10 +34,24 @@ class Monitoring(http.Controller):
     @http.route('/monitoring/status', type='http', auth='none')
     def status(self):
         ensure_db()
+        http_status = 200
         # TODO: add 'sub-systems' status and infos:
         # queue job, cron, database, ...
```
yvaucher, line 39: I guess better to keep database from this todo list, as this is about status and not infos ?

yvaucher (approved):
LGTM

simahawk (approved):

guewen:
`ensure_db()` already runs a sql query

I disabled the connections:

```
# update pg_database set datallowconn = false where datname = 'odoodb';
```

Postgresql logs:

```
db_1     | 2019-10-28 09:22:52 UTC [64]: [1-1] LOG:  connection received: host=192.168.0.3 port=51634
db_1     | 2019-10-28 09:22:52 UTC [64]: [2-1] LOG:  connection authorized: user=odoo database=postgres
db_1     | 2019-10-28 09:22:52 UTC [64]: [3-1] LOG:  duration: 0.073 ms  statement: BEGIN ISOLATION LEVEL REPEATABLE READ
db_1     | 2019-10-28 09:22:52 UTC [64]: [4-1] LOG:  duration: 1.362 ms  statement: select datname from pg_database where datdba=(select usesysid from pg_user where usename='odoo') and not datistemplate and datallowconn and datname not in ('template1', 'postgres') order by datname
db_1     | 2019-10-28 09:22:52 UTC [64]: [5-1] LOG:  duration: 0.033 ms  statement: ROLLBACK
```

Result:

```
$ curl -v http://localhost/monitoring/status                                                                                                                                                 
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /monitoring/status HTTP/1.1
> Host: localhost
> User-Agent: curl/7.47.0
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 404 NOT FOUND
< Content-Type: text/html
< Content-Length: 233
< Server: Werkzeug/0.11.11 Python/2.7.9
< Date: Mon, 28 Oct 2019 09:22:52 GMT
<
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.</p>
* Closing connection 0
```

So I don't think we should run an additional sql query on each healthcheck.

Well, on the other side, the returned error is probably more clear with this change since it returns the exception message.

simahawk:
> Well, on the other side, the returned error is probably more clear with this change since it returns the exception message.

You are right but this is the good part IMO :)

guewen:
> > Well, on the other side, the returned error is probably more clear with this change since it returns the exception message.
> 
> You are right but this is the good part IMO :)

Yes but no, in fact we don't even reach this part, as the dispatcher will already return a 404 error even before calling the route's controller (tested locally, a pdb in the `status` endpoint is never hit when the db is out).

So either we find a way to bypass the normal dispatcher and check "manually" the state of the db, redis, ..., either we rollback these changes. I'd go with the latter since I think the current checks are good enough.

simahawk:
dammit, you are right. The router is going to deal with it earlier.
Do you think we can play w/ `nodb_routing_map`? Server wide modules rely on it so we could make this a server wide module.
I agree current fix does not add anything but does not hurt neither. Let's discuss if we can find an easy way. Otherwise we can roll it back later.

guewen:
> Do you think we can play w/ `nodb_routing_map`? Server wide modules rely on it so we could make this a server wide module.

Good point! When the module is in server wide modules, we reach the method even when there is no database -> we can then test the db status. 
That's a broad change though.
With the current code, the call to `ensure_db()` redirects to the `/web/database/selector` url with a 303. We'll have at least to:

* remove `ensure_db()` I guess
* call again the db list (we cannot execute the current query that would raise `RuntimeError: request not bound to a database` if `ensure_db()` is not called before)? Should we check that we can connect on postgres or that we can connect to individual databases?
* ensure that it runs both with the addon in server wide modules or not (maybe with a warning in the second case) to ease the transition in existing projects

For the record, I tested that we hit the session storage (redis) in any situations, and this is true as long as we return a `Result` (but not with `ensure_db()` which returns before saving the session).

Still, do we really need it? The healthcheck already has enough information (the http status code > 399), adding an extra SQL query for each healthcheck is not that much, but multiply this by the number of health-checkers (3 hosts per container) * number of containers * number projects * 2-3 seconds, it starts to be more... Plus the effort to switch. I'd say, it's useful only if it paves the way for more checks. Really the thing we would need is the health checker being able to tell if a container is out of service because of DB, redis connection, ... or does not respond because all workers are busy with a large task (report, ...). I don't think this change would help here.

simahawk:
I see let's roll it back

PR #106, opened 2019-10-29
Revert "[13.0] [IMP] monitoring_status: check the db is reachable"

guewen:
Reverts camptocamp/odoo-cloud-platform#101

See discussion on camptocamp/odoo-cloud-platform#98

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 cache: pip
 
 python:
-  - "3.5"
+  - "3.6"
 
 addons:
   postgresql: "9.5"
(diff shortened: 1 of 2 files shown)


simahawk:
:red_circle:

guewen:
The error is `Outdated python version detected, Odoo requires Python >= 3.6 to run.` which is clearly unrelated. Restarting the job...

guewen:
Merging because it's a revert and the fail is unrelated: it fails on 13.0 too.
Now by using python3.6 in the build, it goes farther, but the build fails because `cloud_platform_ovh` and `attachment_swift` are not installlable.

PR #108, opened 2019-11-12
[WIP]  9.0 improvement

sebastienbeau:
@simahawk @guewen 
Hi I have done some (lot) refactor in order to have more information.
- having the detail for http request
- having data about time and number of sql request (slow and normal request)
- having data for the cron (state, timing)

I have done the PR in 9 because is the older version and I need to make it working on 8 (the module 9 work perfectly on 8).

I will totally understand if you do not want to merge this change as there broke the current version.
The aim of my PR is more to show you what I am working on and get your feedback.

TODO
-  [ ] metric for queue_job (I think I will add an extra module)

### DIFF
--- a/monitoring_statsd/models/ir_http.py
+++ b/monitoring_statsd/models/ir_http.py
@@ -6,6 +6,7 @@ from openerp import models
 from openerp.http import request
 from datetime import datetime
 from ..statsd_client import statsd, customer, environment
+from ..sql_tracker import get_cursor_tracker
 
 
 class IrHttp(models.AbstractModel):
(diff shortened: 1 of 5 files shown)


simahawk:
@sebastienbeau what about a README and an example of output? :)

guewen:
Very interesting! You might want to have a look at https://github.com/camptocamp/odoo-cloud-platform/pull/37 which is close in the goal (with a different tooling though).

Out of curiosity, did you benchmark the overhead, especially on `Cursor.execute` which is called a lot?

monitoring_statsd/models/ir_http.py
```diff
@@ -16,26 +22,37 @@ def _dispatch(self):
...
+        name = '.'.join([
+            path_info.replace('.', '-'),
+            customer,
+            environment,
+            request.params.get('model', 'undefined').replace('.', '_'),
+            action
+            ])
+
+        with statsd.pipeline() as pipe:
+            start = datetime.now()
```
guewen, line 50: I would use `time.time()`

monitoring_statsd/sql_tracker.py
```diff
@@ -0,0 +1,75 @@
...
+            if surfixe:
+                prefix = "{}_{}".format(prefix, surfixe)
+            if name:
+                return "{}.{}".format(prefix, name)
+            else:
+                return prefix
+
+        pipe.timing(metric_name(prefix, None, name), self.duration)
+        pipe.incr(metric_name(prefix, 'total', name), self.count)
+        if self.slow_active:
```
guewen, line 46: I'm not sure how `slow_active` is meant to be used

monitoring_statsd/sql_tracker.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+ori_execute = Cursor.execute
+
+
+def execute(self, query, params=None, log_exceptions=None):
+    start = datetime.now()
+    res = ori_execute(
+        self, query, params=params, log_exceptions=log_exceptions)
+    end = datetime.now()
```
guewen, line 67: I would use `time.time()`, again ;)

monitoring_statsd/sql_tracker.py
guewen: Ok, got it :)

PR #109, opened 2019-11-14
[12.0] attachment_s3: allow to override connection parameters

grindtildeath:
_get_s3_connection_params can now be overrided to use different connection parameters

### DIFF
--- a/attachment_s3/models/ir_attachment.py
+++ b/attachment_s3/models/ir_attachment.py
@@ -32,6 +32,33 @@ class IrAttachment(models.Model):
         return l
 
     @api.model
+    def _get_s3_connection_params(self, bucket_name=None):
+        host = os.environ.get('AWS_HOST')
+
+        # Ensure host is prefixed with a scheme (use https as default)
+        if host and not urlsplit(host).scheme:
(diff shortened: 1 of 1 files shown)


simahawk (approved):
LG

attachment_s3/models/ir_attachment.py
```diff
@@ -78,6 +66,25 @@ def _get_s3_bucket(self, name=None):
                     ) % (bucket_name, bucket_name)
 
             raise exceptions.UserError(msg)
```
guewen: Shouldn't we raise this exception in _get_s3_bucket after we call `_get_s3_connection_params()`?
It allows a chance to call `super()` and adjust the parameters afterwards without getting an exception.

simahawk: hmm you make a good point. I agree.

simahawk (changes requested):
@grindtildeath pls check Guewen's comment

grindtildeath:
@simahawk Sorry for the delay :sweat_smile: 

Fixed and rebased, if you want to take a look :wink:

simahawk (approved):

PR #110, opened 2019-11-18
[12.0] Add support of 'labs' RUNNING_ENV in cloud_platform

guewen:
The labs env can be anything starting by 'labs', such as
'labs-logistics', 'labs-finance', ...

* At install, s3/swift is set as default storage
* However, unlike prod/integration, the storage is not forced to be an
object storage
* Redis is required
* When the storage is set on s3/swift, then the bucket name is mandatory
(otherwise, there is no place where to create the files...)

The redis prefix regex match is relaxed: anything starting by a project
name, then '-odoo-', then any combination of letters, digits, and dashes
is accepted (so a prefix my-project9-odoo-labs-web3 is valid).


p-tombez (approved):

guewen:
Tested on a labs

PR #115, opened 2019-12-03
[IMP] route file to db base on size and mimetype

vrenaville:
* Purpose
We need to speed up some parts of the storage of image on plateform, we need to avoid too much interaction with external storage service, if image (or other file mimetype ) are small we will store them on DB
* dev
I check the mimetype from a list , and also the filesize, if both check match I store it on DB otherwise file will landed on external storage

### DIFF
--- a/base_attachment_object_storage/data/res_config_settings_data.xml
+++ b/base_attachment_object_storage/data/res_config_settings_data.xml
@@ -1,0 +1,16 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo noupdate="1">
+    <record id="default_mimes_type_storedb" model="ir.config_parameter">
+        <field name="key">mimetypes.list.storedb</field>
+        <field name="value">image</field>
+    </record>
+    <record id="default_file_maxsize_storedb" model="ir.config_parameter">
+        <field name="key">file.maxsize.storedb</field>
+        <field name="value">50000</field>
+    </record>
+    <record id="excluded_model_storedb" model="ir.config_parameter">
+        <field name="key">excluded.models.storedb</field>
+        <field name="value">mail.message,mail.mail</field>
+    </record>
+
+</odoo>
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -10,6 +10,7 @@ import odoo
 
 from contextlib import closing, contextmanager
 from odoo import api, exceptions, models, _
+from odoo.tools.mimetypes import guess_mimetype
 
 
 _logger = logging.getLogger(__name__)
@@ -36,8 +37,6 @@ def clean_fs(files):
 
 class IrAttachment(models.Model):
     _inherit = 'ir.attachment'
-
-    _local_fields = ('image_small', 'image_medium', 'web_icon_data')
 
     def _register_hook(self):
         super()._register_hook()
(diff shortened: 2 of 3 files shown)


base_attachment_object_storage/data/res_config_settings_data.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='utf-8'?>
+<odoo noupdate="1">
+    <record id="default_mimes_type_storedb" model="ir.config_parameter">
+        <field name="key">mimetypes.list.storedb</field>
+        <field name="value">image</field>
+    </record>
+    <record id="default_file_maxsize_storedb" model="ir.config_parameter">
+        <field name="key">file.maxsize.storedb</field>
+        <field name="value">25000</field>
```
guewen: I would put 50000, as I found `image_medium` of this size in the existing databases, and we want all the `image_medium` equivalent images to be stored in database (they can be used in kanban views for instance).

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -89,12 +90,19 @@ def _save_in_db_anyway(self):
             return True
 
         # Binary fields
-        if self.res_field:
-            # Binary fields are stored with the name of the field in
-            # 'res_field'
-            # 'image' fields can be rather large and should usually
-            # not be requests in bulk in lists
-            if self.res_field and self.res_field in self._local_fields:
```
guewen, line 97: `_local_fields` can be removed from the class

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -89,12 +90,19 @@ def _save_in_db_anyway(self):
...
+        data_to_store = self.datas
+        bin_data = base64.b64decode(data_to_store) if data_to_store else ''
+        current_mimetype = guess_mimetype(bin_data)
+        mimetypes_settings = self.env['ir.config_parameter'].sudo().get_param(
+            'mimetypes.list.storedb')
+        mimetypes_for_db_store = mimetypes_settings \
+            and mimetypes_settings.split(',') or []
```
guewen: Avoid ` ... and ... or ...`.
`IrConfigParameter.get_param()` accepts a `default` argument, so you should be able to do:

```
        mimetypes_settings = self.env['ir.config_parameter'].sudo().get_param(
            'mimetypes.list.storedb', default=''
        )
        mimetypes_for_db_store = mimetypes_settings.split(',')
```

guewen (changes requested):
A few details but looks good! thanks

guewen (changes requested):
A few details but looks good overall. Thanks!

guewen (approved):

p-tombez (commented):
LG
I'm just a bit concerned if we could like having models excluded from this check.
e.g. Mail objects : some people like to use images in their mail signature (I know it's bad, but it happens way more often than it should...). As images are stored as attachments, with this code, all the images from the signatures will be stored in DB. Depending on the mail volume, it could become an issue in long term

vrenaville:
@p-tombez thanks for your idea , it's a great one
I have adapt the code
@guewen could you please update your review

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -87,14 +86,28 @@ def _save_in_db_anyway(self):
...
-
-        # Binary fields
-        if self.res_field:
-            # Binary fields are stored with the name of the field in
-            # 'res_field'
-            # 'image' fields can be rather large and should usually
-            # not be requests in bulk in lists
-            if self.res_field and self.res_field in self._local_fields:
+        import pdb
+        pdb.set_trace()
```
guewen: 🙊

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -87,14 +86,28 @@ def _save_in_db_anyway(self):
...
+        # Check if model must never be stored on DB
+        excluded_model_settings = self.env['ir.config_parameter'].sudo().\
+            get_param('excluded.models.storedb', default='')
+        excluded_model_for_db_store = excluded_model_settings.split(',')
+        if any(self.res_model == val for val in
+               excluded_model_for_db_store):
```
guewen: `if self.res_model in excluded_model_for_db_store:`?

vrenaville: yep much better

vrenaville: Friday copy/paste :) , I have update it

guewen (approved):

p-tombez (approved):

PR #121, opened 2019-12-12
[12.0] base_fileurl_field: add acl_public_read slot

grindtildeath:


base_fileurl_field/fields.py
```diff
@@ -33,6 +33,7 @@ class FileURL(fields.Binary):
         'storage_location': '',  # External storage activated on the system (cf base_attachment_storage)  # noqa
         'storage_path': '',  # Path to be used as storage key (prefix of filename)  # noqa
         'filename': '',  # Field to use to store the filename on ir.attachment
+        'acl_public_read': True,  # Set public-read ACL by default
```
guewen, line 36: There is a `public` boolean on `ir.attachment`. AFAIU, when the `public` field is True, attachments bypass the ACL checks.
Wouldn't it make sense to set the `public` field to True then? That would solve another issue which is that currently you need to re-open a connection to s3 in `_store_file_set_acl()`. If we use the `public` field, we could use the same connection just after we push the file to set the S3 ACL.

Tonow-c2c (commented):
#122 should not be in same PR ?

base_fileurl_field/fields.py
grindtildeath: Sure it would make sense to do both operations (create + set ACL) in the same s3 connection.

However, as `_store_file_write` is `api.model`, we don't have a link to the record anymore in order to check `public` value, so we'd need to use a context key to update ACL accordingly.

But, if we rely only on `public`, we'd then need to override `create` and `write` on `ir.attachment` to update s3 permissions, when `public` is updated. Then, the issue is that updates on s3 are triggered by `_inverse_datas`, which is not triggered when changing 'only' public. That means, we'd need to add a proxy function in `base_attachment_object_storage` and call this one, when only public is updated...

I'm feeling like it gets too complicated for something that should happen only once (at record creation) and that we can simply manage with a second call to the s3 api. 

Therefore I just added the `public` key in the vals dict to create the `ir.attachment`, so that it's "publicly available" from Odoo as long as it is "publicly available" from s3.

guewen: Sure, it makes sense. Thanks for writing notes about this.

p-tombez (approved):

grindtildeath:
PR closed as we move from ACL per file to bucket policies using prefixes.

PR #79, opened 2019-08-09
[12.0] Add method to force storage of special attachments to DB

sebalix:
Port of #54 to v12.

PR #123, opened 2019-12-12
[13.0] Add method to force storage of special attachments to DB

vrenaville:
[13.0] Add method to force storage of special attachments to DB

### DIFF
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -5,12 +5,15 @@ import base64
 import inspect
 import logging
 import os
+import time
+
 import psycopg2
 import odoo
 
 from contextlib import closing, contextmanager
 from odoo import api, exceptions, models, _
 from odoo.tools.mimetypes import guess_mimetype
+from odoo.osv.expression import AND, normalize_domain
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


guewen:
You miss the `_save_in_db_domain` from https://github.com/camptocamp/odoo-cloud-platform/pull/79/files#diff-b57e51260befc18fdcdafef0b725cef2R73-R92 which reflects the same conditions as you have in https://github.com/camptocamp/odoo-cloud-platform/blob/aa85eeb4bccdf9864a467ccf57ae9fee4da6244b/base_attachment_object_storage/models/ir_attachment.py#L72-L118 but in the form of a domain.

In the idea of (but with proper values from the config):

```python

domain = [
            '|',
            # assets are stored in 'ir.ui.view'
            ('res_model', '=', 'ir.ui.view'),
            '&', '&',
            ('file_size', '<', 50000),
            ('res_model', 'not in', ('xxx')),
            ('mimetype', '=like', 'image/%'),
        ]
```

vrenaville:
@guewen 
Yes I have seen yesterday but I was on an hurry, I take time to finish it today.
I have made a test locally and it seems to work well

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -65,6 +68,34 @@ def _register_hook(self):
...
+            'file.maxsize.storedb', default='0')
+
+        return [
+            '|',
+            # assets are stored in 'ir.ui.view'
+            ('res_model', '=', 'ir.ui.view'),
+            '&', '&',
+            ('file_size', '<', int(filesize)),
+            ('res_model', 'not in', excluded_model_for_db_store),
+            ('mimetype', '=like', mimetypes_for_db_store),
```
guewen: Since you can have more than one mimetype, and we are looking for the beginning only...
`not in` can't work, `=like` on a list neither. Should we add one tuple per possible mimetype separated by `|`?

Would that work?
```
domain += ['|'] * len(mimetypes_for_db_store) - 1
domain += [('mimetype', '=like', mimetype) for mimetype in mimetypes_for_db_store]
```

vrenaville: @guewen thanks for pointing it, I have done a fix, I have try both cases (no mimetypes + several ones), and it seems to works

guewen (approved):
Looks good! Can you squash your 2 commits?

p-tombez (approved):

PR #124, opened 2019-12-13
[7.0] Fix session expiration ignored

p-tombez:
Fix issue when expiration set on the session is not taken in consideration when saving to Redis

### DIFF
--- a/session_redis/session.py
+++ b/session_redis/session.py
@@ -38,14 +38,20 @@ class RedisSessionStore(SessionStore):
 
     def save(self, session):
         key = self.build_key(session.sid)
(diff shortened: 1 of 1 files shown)


session_redis/session.py
```diff
@@ -38,14 +38,16 @@ def build_key(self, sid):
 
     def save(self, session):
         key = self.build_key(session.sid)
+        session_oe = session.itervalues().next()
```
guewen, line 45: What's the type of session? :thinking: Could you add a comment explaining why we have to look for the first value?

guewen (approved):

Tonow-c2c (approved):
thanks :pray:

PR #127, opened 2020-01-14
Store .less files in database

guewen:
When users customize .less files using the 'web_editor', the controller
creates a new ir.attachment for the customized file [0].

The attachment will be stored in S3 by default, which is unfortunate since it will
slow down the generation of the assets. Force storage in database for .less files.

Note: beware when porting this to upper versions, check the code of
web_editor, because it changed (notably less files are now scss).

[0] https://github.com/odoo/odoo/blob/11.0/addons/web_editor/controllers/main.py#L370

### DIFF
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -83,6 +83,12 @@ class IrAttachment(models.Model):
             '|',
             # assets are stored in 'ir.ui.view'
             ('res_model', '=', 'ir.ui.view'),
+            # less files customized through 'web_editor' are saved as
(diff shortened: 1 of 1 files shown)


simahawk (approved):
Seems legit. 
Related: I wonder if we can store compiled CSS and JS assets into object storage and serve them straight from it (via public URL). This would avoid tons of requests on Odoo.
I'm not sure is going to be easy to override the URL the asset machinery produces but we could make a rewrite rule on nginx.
The alternative is: put a cache in front of it but it could be troublesome to handle invalidation.

nbessi (approved):

PR #116, opened 2019-12-05
Add AWS_BUCKETNAME_UNSTRUCTURED to by-pass check

yvaucher:
Curently the name of the bucket is cross checked with the
running environment. In rare case you can have a bucket name
that doesn't match the structure `<project>-odoo-<env>` in place.

PR #141, opened 2020-02-05
[12.0] Require AWS_BUCKETNAME for prod and integration environments only

grindtildeath:

Fixes bug introduced by 6c3b610

### DIFF
--- a/cloud_platform/models/cloud_platform.py
+++ b/cloud_platform/models/cloud_platform.py
@@ -100,19 +100,18 @@ class CloudPlatform(models.AbstractModel):
                 "SWIFT_PASSWORD environment variable is required when "
                 "ir_attachment.location is 'swift'."
             )
-            container_name = os.environ.get('SWIFT_WRITE_CONTAINER')
-            assert container_name, (
-                "SWIFT_WRITE_CONTAINER environment variable is required when "
-                "ir_attachment.location is 'swift'.\n"
-                "Normally, 'swift' is activated on labs, integration "
-                "and production, but should not be used in dev environment"
-                " (or using a dedicated dev bucket, never using the "
-                "integration/prod bucket).\n"
-                "If you don't actually need a bucket, change the"
-                " 'ir_attachment.location' parameter."
-            )
-            prod_container = bool(re.match(r'[a-z0-9-]+-odoo-prod',
-                                           container_name))
+            container_name = os.environ.get('SWIFT_WRITE_CONTAINER', '')
+            if environment_name in ('prod', 'integration', 'labs'):
+                assert container_name, (
+                    "SWIFT_WRITE_CONTAINER environment variable is required when "
+                    "ir_attachment.location is 'swift'.\n"
+                    "Normally, 'swift' is activated on labs, integration "
+                    "and production, but should not be used in dev environment"
+                    " (or using a dedicated dev bucket, never using the "
+                    "integration/prod bucket).\n"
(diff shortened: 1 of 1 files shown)


cloud_platform/models/cloud_platform.py
```diff
@@ -156,7 +156,10 @@ def _check_s3(self, environment_name):
                 "AWS_SECRET_ACCESS_KEY environment variable is required when "
                 "ir_attachment.location is 's3'."
             )
-            bucket_name = os.environ.get('AWS_BUCKETNAME')
+            bucket_name = (
+                os.environ.get('AWS_BUCKETNAME') or
+                os.environ.get('ODOO_DEV_KEEP_ATTACHMENT_S3')
```
simahawk: what about `AWS_BUCKET_READONLY`?
Also, shall we document it somewhere?

Tonow-c2c (approved):
LGTM with answer about @simahawk question

cloud_platform/models/cloud_platform.py
jcoux: To be consistent, we need to use a common variable for S3 and SWIFT.
The modification here must be added also into function `_check_swift` just above this function.
And the name of environment variable must be changed into odoo-template:  (EDIT: private link removed (AJU))

jcoux (changes requested):

grindtildeath:
@simahawk Kind of agree with @jcoux , would make sense to use only one var name for both s3 and swift.

So what do we want to use ? 
1. ODOO_DEV_KEEP_ATTACHMENT_S3 (easiest change since it's the same var used on odoo-template and the provided solution works. We'd still need to add the same `or` condition for swift but the name is then misleading.)
2. ODOO_DEV_KEEP_ATTACHMENT (it only states we keep attachments whatever the storage that is used use and the provided solution works. We'd still need to add the same `or` condition for swift and do the change on odoo-template + all the projects)
3. STORAGE_READONLY (if we want to set such a var, we must ensure _store_file_write and _store_file_delete are overriden to make sure we're in a readonly mode, Might make the setting less obvious since we must handle the case where both the bucketname and this var are set to make sure it won't have side effects. Moreover we'd still need to add the same `or` condition for swift and do the change on odoo-template + all the projects)

Personally I'd favor option 2...

guewen (changes requested):
It looks like a leak of our project's template into this module. I don't think we should have such variable at all. 
Actually, it's a mistake I introduced lately in: https://github.com/camptocamp/odoo-cloud-platform/commit/6c3b610610d7a1c1bc5e51c57840bc67d976dcb3

We used to be able to start odoo using aws accesses but no bucket name before my PR.

The correct solution IMO is to make this check only on a selection of environments (prod, integration, labs) and not introduce a custom variable.

grindtildeath:
@guewen I reworked the PR according to your last comment. Still, as I understand, we don't want to run this check on `labs` ?

guewen (approved):

guewen:
> Still, as I understand, we don't want to run this check on `labs` ?

Discussed this directly but for the record, we want to check this. (using a different `ir_attachment.location` for labs is allowed, but if the value is `s3` or `swift`, the lab will run into issues if the bucket name is not set).

yvaucher:
This is conflicting due to #116 that I just merged. (it was open since a while)

Tell me if you need help to resolve the conflict.

jcoux (approved):
Thanks for changes.

PR #142, opened 2020-02-13
[13.0] Use keystoneauth v3 for Swift attachments

smussie:
[BSCLOUDFR-6] : Changed v2 to v3 and tenant_name to project_name

### DIFF
--- a/attachment_swift/models/ir_attachment.py
+++ b/attachment_swift/models/ir_attachment.py
@@ -46,18 +46,18 @@ class SwiftSessionStore(object):
     def __init__(self):
         self._sessions = {}
 
-    def _get_key(self, auth_url, username, password, tenant_name):
-        return (auth_url, username, password, tenant_name)
+    def _get_key(self, auth_url, username, password, project_name):
+        return (auth_url, username, password, project_name)
 
     def get_session(self, auth_url=None, username=None, password=None,
-                    tenant_name=None):
-        key = self._get_key(auth_url, username, password, tenant_name)
+                    project_name=None):
+        key = self._get_key(auth_url, username, password, project_name)
         session = self._sessions.get(key)
         if not session:
-            auth = keystoneauth1.identity.v2.Password(
+            auth = keystoneauth1.identity.v3.Password(
                 username=username,
                 password=password,
-                tenant_name=tenant_name,
+                project_name=project_name,
                 auth_url=auth_url,
             )
             session = keystoneauth1.session.Session(
@@ -85,12 +85,18 @@ class IrAttachment(models.Model):
         host = os.environ.get('SWIFT_AUTH_URL')
         account = os.environ.get('SWIFT_ACCOUNT')
         password = os.environ.get('SWIFT_PASSWORD')
-        tenant_name = os.environ.get('SWIFT_TENANT_NAME')
+        project_name = os.environ.get('SWIFT_PROJECT_NAME')
+        if not project_name and os.environ.get('SWIFT_TENANT_NAME'):
+            project_name = os.environ['SWIFT_TENANT_NAME']
+            _logger.warning(
+                "SWIFT_TENANT_NAME is deprecated and "
+                "must be replaced by SWIFT_PROJECT_NAME"
+            )
         region = os.environ.get('SWIFT_REGION_NAME')
         os_options = {}
         if region:
             os_options['region_name'] = region
-        if not (host and account and password and tenant_name):
+        if not (host and account and password and project_name):
             raise exceptions.UserError(_(
                 "Problem connecting to Swift store, are the env variables "
                 "(SWIFT_AUTH_URL, SWIFT_ACCOUNT, SWIFT_PASSWORD, "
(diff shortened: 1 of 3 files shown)


attachment_swift/models/ir_attachment.py
```diff
@@ -85,22 +85,22 @@ def _get_swift_connection(self):
         host = os.environ.get('SWIFT_AUTH_URL')
         account = os.environ.get('SWIFT_ACCOUNT')
         password = os.environ.get('SWIFT_PASSWORD')
-        tenant_name = os.environ.get('SWIFT_TENANT_NAME')
+        project_name = os.environ.get('SWIFT_PROJECT_NAME')
```
guewen, line 88: To ease the transition and prevent some mistakes on deployments, I would still support the old name `SWIFT_TENANT_NAME`, with a warning. Example:

```python
project_name = os.environ.get('SWIFT_PROJECT_NAME')
if not project_name and os.environ.get('SWIFT_TENANT_NAME'):
    project_name = os.environ['SWIFT_TENANT_NAME']
    _logger.warning(
        "SWIFT_TENANT_NAME is deprecated and "
        "must be replaced by SWIFT_PROJECT_NAME"
    )
```

guewen (commented):
I didn't read the swift specs but the change looks good.
I have a suggestion for backward compatibility.
Can you check if there is anything to change in the readme file of the module

Thanks!

attachment_swift/models/ir_attachment.py
```diff
@@ -85,12 +85,16 @@ def _get_swift_connection(self):
         host = os.environ.get('SWIFT_AUTH_URL')
         account = os.environ.get('SWIFT_ACCOUNT')
         password = os.environ.get('SWIFT_PASSWORD')
-        tenant_name = os.environ.get('SWIFT_TENANT_NAME')
+        project_name = os.environ.get('SWIFT_TENANT_NAME'):
+        _logger.warning(
+            "SWIFT_TENANT_NAME is deprecated and "
+            "must be replaced by SWIFT_PROJECT_NAME"
+        )
```
guewen: Hmm but you are not using `SWIFT_PROJECT_NAME` anymore. My proposal was:

* if SWIFT_PROJECT_NAME is set: use it (that's what we'll want at the end)
* but if SWIFT_PROJECT_NAME is not set and SWIFT_TENANT_NAME is set, it means people are still using the old name for the variable: use it so nothing is broken, but log a message so people know they have to change it

Happy to discuss about this with you on a call if you need

attachment_swift/tests/test_mock_swift_api.py
```diff
@@ -67,10 +69,16 @@ def test_connection(self, mock_swift_client):
...
-        self.assertEqual(session.auth.username, os.environ['SWIFT_ACCOUNT'])
-        self.assertEqual(session.auth.password, os.environ['SWIFT_PASSWORD'])
-        self.assertEqual(session.auth.tenant_name,
-                         os.environ['SWIFT_TENANT_NAME'])
+        self.assertEqual(session.auth.get_cache_id_elements().get(
+            'password_username'), os.environ['SWIFT_ACCOUNT'])
+        self.assertEqual(session.auth.get_cache_id_elements().get(
+            'password_password'), os.environ['SWIFT_PASSWORD'])
+        self.assertEqual(session.auth.project_name,
+                         os.environ['SWIFT_TENANT_NAME']):
```
guewen: looks like a copy-paste ;)

smussie: Sorry, I didn't understand your proposal at first because i am null in python, but i will correct it now. And thanks for your review and patience

smussie: Sorry to bother, but what do you think now?

guewen:
Great! It looks good :) One remaining thing to do: update the documentation in the readme file.

vrenaville (approved):

smussie:
@guewen I didn't find anything that is directly influnced by the changes made in the readme file, thus in my humble opinion I left it as is, but if you find that there is can you please point it out to me?

guewen:
On my phone now but do a grep of SWIFT_TENANT_NAME in the repository and you will find an occurrence in README.rst (always a good practice to find and check every occurrence of a string with grep/ack-grep, the find function of your editor when renaming or deleting something). 

edit: also this readme file contains a section explaining how to test manually, and the example uses v2, which makes no sense now.

Tonow-c2c (commented):
@smussie 
like say @guewen you still have reference to 
```
attachment_swift/README.rst
16:* ``SWIFT_TENANT_NAME``
47:    export OS_TENANT_NAME={SWIFT_TENANT_NAME}
```

p-tombez:
@Tonow-c2c @guewen Doc has been updated

guewen (approved):
Thanks! The last paragraph's rendering is broken (it was already), would be worth fixing, but change is good anyway.

Tonow-c2c (approved):

PR #147, opened 2020-02-25
[13.0]Require AWS_BUCKETNAME for prod, integration and labs environments only

grindtildeath:

Forward port of #141


yvaucher:
You might want to rebase on #151 in order to keep the same merge order as on branch 12.0

grindtildeath:
Fast-tracking

PR #150, opened 2020-02-25
[9.0] Require AWS_BUCKETNAME for prod, integration and labs environments only

grindtildeath:

Backport of #141


grindtildeath:
Fast-tracking

PR #176, opened 2020-04-03
Converting date/datetime python objects to string before serializing

PCatinean:
It appears that odoo enterprise module website_sale_coupon (or potentially other modules for that matter) use python objects to store session data `request.session['last_coupon_update'] = datetime.utcnow()`. While this is definitely not good practice it does error out when attempting to serialize.

This fix adds conversion for Date/Datetime objects and leaves the method to be extended for other objects.

Thanks for sharing this great module :)

### DIFF
--- a/session_redis/session.py
+++ b/session_redis/session.py
@@ -20,6 +20,9 @@ _logger = logging.getLogger(__name__)
 class RedisSessionStore(SessionStore):
     """ SessionStore that saves session to redis """
 
+    py_date_prefix = '__py__date_'
+    py_datetime_prefix = '__py__datetime_'
+
     def __init__(self, redis, session_class=None,
                  prefix='', expiration=None, anon_expiration=None):
         super(RedisSessionStore, self).__init__(session_class=session_class)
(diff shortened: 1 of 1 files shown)


simahawk:
Ciao caro :)
There are several places where they do it wrong. We fixed this recently https://github.com/odoo/odoo/pull/40850/files
Where is this variable used? Can't find it in v12...

In any case, I don't think is good to change the way it's stored in the session -> if the application expects it to be a date you'll break that behavior, unless you convert it back to the same object on lookup. It would be better to fix it in odoo core IMO.
Am I wrong?

PCatinean:
Ciao Simone <3

Yes you might be right on this one, I think my initial train of thought was to make it work as the default session storage. Clearly this is not it but I can't help but wonder how it's done. Maybe they just store the datetime object as string and it's somehow reinterpreted when retrieved?

The code I pasted is from enterprise addons webiste_sale_coupon/controllers/main.py.

PCatinean:
So as it turns out, after a bit of investigation, the standard werkzeug session storage class uses the pickle library to store and retrieve actual python objects.

This means that not only the date/datetime is the issue here but any parts of the code that might store a specific python object.

Given this new finding would it be reasonable to assume that storing data with pickle would be recommended as this is what the Odoo framework assumes?

If not would it be useful to at least attempt to support date/datetime objects by storing them with a prefix before the key and transform them before retrieval?

I'm curious as how you did not have issues in production so far as the module I mentioned is standard enterprise and error-ed out when I visited the chart in the webshop.

guewen:
Hey @PCatinean :)

We may want to support a custom JSON serializer/deserializer to support at least datetime objects (examples in https://github.com/OCA/queue/blob/dc12a6a20ecfd15c5b90f9b089c9a64cf9d8bbe4/queue_job/fields.py#L57-L99). Which is eventually equivalent but a bit cleaner?

> I'm curious as how you did not have issues in production so far as the module I mentioned is standard enterprise and error-ed out when I visited the chart in the webshop.

I guess we don't have that much webshop instances (and the ones I know, we applied the patch mentionned by @simahawk and probably didn't use sale coupons)

simahawk:
@PCatinean sorry, I forgot to get back to you... puoi perdonarmi? :smile: 
I agree w/ Guewen: go for custom enc/dec classes. 
And yes, we were lucky because we don't have so many instances w/ apps writing a lot on session or doing it properly :man_shrugging:

PCatinean:
Hi guys! I see then it makes sense for your use case. I noticed this issue in my instance with the enterprise addons and with others modules as well. 

My train of thought was more like -> If any addon whatsoever tries to store data in the session that's not json serializable the module should at the very least throw a warning so you know there are issues and at the second level try to handle it gracefully so you don't end up with with a crashed instance :).

How do you think we should proceed in this case?

PCatinean:
Closing in favor of https://github.com/camptocamp/odoo-cloud-platform/pull/178

PR #177, opened 2020-04-15
11.0 session redis preserve sessions

ap-wtioit:
When testing the session redis store on one of our instances we didn't want to destroy 2.6 million sessions.

With making the purge sessions feature optional we prevent purging all the sessions (that are still used by other odoo instances with the same session dir) when just testing it works.

Also added a copy odoo session feature that allows copying existing sessions to redis before restarting odoo with redis session store in effect. This helps reducing the number of sessions lost when switching to redis.

Info @wt-io-it

### DIFF
--- a/session_redis/README.rst
+++ b/session_redis/README.rst
@@ -18,6 +18,10 @@ The storage of sessions in Redis is activated using environment variables.
   the sessions (default is 7 days)
 * ``ODOO_SESSION_REDIS_EXPIRATION_ANONYMOUS`` is the time in seconds before expiration of
   the anonymous sessions (default is 3 hours)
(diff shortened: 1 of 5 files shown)


guewen:
Makes sense, nice addition.
Could you fix the lint please?

Reported in travis:
```
session_redis/http.py:82:89: E501 line too long (113 > 88 characters)
session_redis/http.py:84:20: E225 missing whitespace around operator
session_redis/http.py:89:89: E501 line too long (99 > 88 characters)
```

guewen:
I'm looking why the build is red, seems to be linked to a change in the module server_environment in OCA/server-env. In any case, not related to your change.

PR #179, opened 2020-05-11
[11.0] Fix build after changes in server_environment

guewen:
Now the default env is "test", and we don't need the directory
at all.
See https://github.com/OCA/server-env/pull/44


Tonow-c2c (commented):
it's only for the branch 11.0 ?

guewen:
> it's only for the branch 11.0 ?

The other builds are not failing

guewen:
It fails on 12.0, going to port this change

PR #183, opened 2020-05-14
[13.0][IMP] session_redis: allow configuring Redis URL instead of host & port.

mlaitinen:
Added a configuration option `ODOO_SESSION_REDIS_URL` to `session_redis` that makes it possible to use Redis behind a secure connection (`rediss://username:password@host:port`).

### DIFF
--- a/session_redis/README.rst
+++ b/session_redis/README.rst
@@ -13,6 +13,8 @@ The storage of sessions in Redis is activated using environment variables.
 * ``ODOO_SESSION_REDIS_PORT`` is the redis port (default is ``6379``)
 * ``ODOO_SESSION_REDIS_PASSWORD`` is the password for the AUTH command
   (optional)
+* ``ODOO_SESSION_REDIS_URL`` is an alternative way to define the Redis server
+  address. It's the preferred way when you're using the ``rediss://`` protocol.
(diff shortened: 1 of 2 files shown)


session_redis/README.rst
```diff
@@ -13,6 +13,8 @@ The storage of sessions in Redis is activated using environment variables.
 * ``ODOO_SESSION_REDIS_PORT`` is the redis port (default is ``6379``)
 * ``ODOO_SESSION_REDIS_PASSWORD`` is the password for the AUTH command
   (optional)
+* ``ODOO_SESSION_REDIS_URL`` is an alternative way to define the Redis server
+  address. It's the preferred way when you're using the ``rediss://`` protocol.
```
guewen, line 17:
```suggestion
  address. It's the preferred way when you're using the ``redis://`` protocol.
```

mlaitinen: `rediss` wasn't a typo. The reason why I added the option to define the address in the form of an URL is because you can't connect to a TLS secured Redis with just the host and port.

guewen: Ok, I wasn't aware of this!

guewen (approved):
Thanks @mlaitinen

PR #186, opened 2020-05-27
[13.0] Fix domain for finding attachments to store in DB

guewen:
'=like' is converted to a normal sql LIKE, as the configuration for
mimetypes is "image" and not "image/png", "image/jpeg", ... we have
to search using a wildcard pattern suffix


vrenaville (approved):

guewen:
Replaced by #186

PR #187, opened 2020-05-27
[13.0] Rework and fix storage forced in database

guewen:
The initial issue that triggered this rework is that the forced storage in
database was working only on writes, and was never applied on attachment
creations.

This feature is used to store small files that need to be read in a fast way in
database rather than in the object storage. Reading a file from the object
storage can take 150-200ms, which is fine for downloading a PDF file or a single
image, but not if you need 40 thumbnails.

Down the path to make a correction, I found that:

* the logic to force storage was called in `_inverse_datas`, which is not called
  during a create
* odoo implemented a new method `_get_datas_related_values`, which is a model
  method that receive only the data and the mimetype, and return the attachment
  values and write the file to the correct place

The `_get_datas_related_values` is where we want to plug this special storage,
as it is called for create and write, and already handle the values and
conditional write. But using this method, we have less information than before
about the attachment, so let's review the different criterias we had before:

* res_model: we were using it to always store attachments related to
  'ir.ui.view' in db, because assets are related to this model. However, we
  don't really need to check this: we should store any javascript and css
  documents in database.
* exclude res_model: we could have an exclusion list, to tell that for instance,
  for mail.message, we should never store any image in db. We don't have this
  information anymore, but I think it was never used and added "in case of".
  Because the default configuration is "mail.mail" and "mail.message" and I
  couldn't find any attachment with such res_model in any of our biggest
  databases. So this is removed.
* mimetype and data (size) are the last criteria and we still have them

The new system is only based on mimetype and data size and I think it's actually
more versatile. Previously, we could set a global size and include mimetypes,
but we couldn't say "I want to store all images below 50KB and all files of type
X below 10KB". Now, we have a single system parameter with a dict configuration
(`ir_attachment.storage.force.database`) defaulting to:

    {"image/": 50000, "application/javascript": 0, "text/css": 0}

Assets have a limit of zero, which means they will all be stored in the database
whatever their size is.

Overall, this is a great simplification of the module too, as the method
`_get_datas_related_values` integrates it better in the base calls of IrAttachment.

Note for upgrade:

I doubt we customized the previous system parameters which are now obsolete, but
if yes, the configuration may need to be moved to `ir_attachment.storage.force.database`.
For the record, the params were:

* mimetypes.list.storedb (default: image)
* file.maxsize.storedb (default: 50000)
* excluded.models.storedb (mail.message,mail.mail), no equivalent now

The method `IrAttachment.force_storage_to_db_for_special_fields()` should be called
through a migration script on existing databases to move the attachments back into
the database.

### DIFF
--- a/base_attachment_object_storage/data/res_config_settings_data.xml
+++ b/base_attachment_object_storage/data/res_config_settings_data.xml
@@ -1,16 +1,9 @@
 <?xml version='1.0' encoding='utf-8'?>
 <odoo noupdate="1">
-    <record id="default_mimes_type_storedb" model="ir.config_parameter">
-        <field name="key">mimetypes.list.storedb</field>
-        <field name="value">image</field>
-    </record>
-    <record id="default_file_maxsize_storedb" model="ir.config_parameter">
-        <field name="key">file.maxsize.storedb</field>
-        <field name="value">50000</field>
-    </record>
-    <record id="excluded_model_storedb" model="ir.config_parameter">
-        <field name="key">excluded.models.storedb</field>
-        <field name="value">mail.message,mail.mail</field>
+
+    <record id="ir_attachment_storage_force_database" model="ir.config_parameter">
+        <field name="key">ir_attachment.storage.force.database</field>
+        <field name="value">{"image/": 51200, "application/javascript": 0, "text/css": 0}</field>
     </record>
 
 </odoo>
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -12,8 +12,8 @@ import odoo
 
 from contextlib import closing, contextmanager
 from odoo import api, exceptions, models, _
-from odoo.tools.mimetypes import guess_mimetype
-from odoo.osv.expression import AND, normalize_domain
+from odoo.osv.expression import AND, OR, normalize_domain
+from odoo.tools.safe_eval import const_eval
 
 
 _logger = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


jgrandguillaume (approved):
Nice one :) !

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -68,109 +68,114 @@ def _register_hook(self):
...
-        ]
-        domain += ['|'] * (len(mimetypes_for_db_store) - 1)
-        domain += [('mimetype', '=like', mimetype) for mimetype in
-                   mimetypes_for_db_store]
+        domain = []
+        storage_config = self._get_storage_force_db_config()
+        for mimetype_key, limit in storage_config.items():
+            part = [("mimetype", "=like", "{}%".format(mimetype_key))]
+            if limit:
+                part = AND([part, [("file_size", "<=", limit)]])
```
grindtildeath, line 110: Here I guess, we must ensure limit is a Integer > 0, but it's probably better to do it `_get_storage_force_db_config` ?

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -68,109 +68,114 @@ def _register_hook(self):
         if update_module:
             self.env['ir.attachment'].sudo()._force_storage_to_object_storage()
 
-    @api.model
-    def _save_in_db_domain(self):
+    @property
+    def _object_storage_default_force_db_config(self):
+        return {"image/": 50000, "application/javascript": 0, "text/css": 0}
```
grindtildeath:
```suggestion
        return {"image/": 51200, "application/javascript": 0, "text/css": 0}
```
Otherwise a 48.9kB attachment will still be stored on object storage. This ensure it will be 50k as stated in the docs.

base_attachment_object_storage/data/res_config_settings_data.xml
```diff
@@ -1,16 +1,9 @@
...
-        <field name="key">file.maxsize.storedb</field>
-        <field name="value">50000</field>
-    </record>
-    <record id="excluded_model_storedb" model="ir.config_parameter">
-        <field name="key">excluded.models.storedb</field>
-        <field name="value">mail.message,mail.mail</field>
+
+    <record id="ir_attachment_storage_force_database" model="ir.config_parameter">
+        <field name="key">ir_attachment.storage.force.database</field>
+        <field name="value">{"image/": 50000, "application/javascript": 0, "text/css": 0}</field>
```
grindtildeath:
```suggestion
        <field name="value">{"image/": 51200, "application/javascript": 0, "text/css": 0}</field>
```
cf below

grindtildeath (commented):
Great improvement :+1: 
Only two small things to pay attention to.

base_attachment_object_storage/models/ir_attachment.py
guewen: I copied the previous configuration, but yes, it's better as you suggest, going to change

base_attachment_object_storage/models/ir_attachment.py
guewen: If someone puts a negative limit, no file will match this mimetype, which means they stay on the object storage, which is equivalent to not put the mimetype in the configuration. It doesn't really make sense to use a negative limit, but it does no harm, so I don't think we should add code to handle this.

grindtildeath (approved):

PR #188, opened 2020-06-18
[10.0] Extend json logger with request uuid and user id

sbidoul:

PR #192, opened 2020-06-22
[11.0][PORT] 188 from 10.0

sbidoul:
Port #188 from 10.0 to 11.0.


guewen (approved):
And thanks for the ports :)

PR #198, opened 2020-09-03
monitoring_log_requests: Improve data output

p-tombez:
* Add environment var to enable/disable requests log
* Add ability to send data to an UDP listener instead of stdout


guewen:
Do we plan to use this module? Currently AFAIK we never use it in favor of https://github.com/camptocamp/odoo-cloud-platform/tree/13.0/monitoring_statsd, so we can find these stats in grafana

nbessi (approved):

nbessi (commented):
The needs are not the same as monitoring statsd we have a lot more infor in the addons that simply time metric and it takes a lot of tinme

nbessi (approved):

santostelmo (approved):

Issue #201, opened 2020-10-28
file is copied to s3 but stays in the local filesystem with the same name


ayushin:
I'm trying to migrate existing filestore setup to google cloud s3 (few issues there solved, will report later)

but for some reason the migration tool is copying files to the s3 but never updates the stored filename so the system keeps looking for the attachment on the local file storage.

that's version 12.0

anybody else saw that?

ayushin:
ps new uploads work just fine, as expected the file name starts with s3

ayushin:
for some super weird reason the new s3 file name does not get commited to the database and all goes well until:

new_env.cr.after('commit', clean) <- and the commit never happens

ayushin:
I solved this with 

https://github.com/camptocamp/odoo-cloud-platform/pull/202

But can somebody knowledgeable explain to me why I am the only one having this issue?

ayushin:
sorry, rookies mistake... did not realise odoo shell is not committing automatically. what a waste of time

PR #202, opened 2020-10-29
12.0 missing commit

ayushin:
Solves https://github.com/camptocamp/odoo-cloud-platform/issues/201 for me.

I don't understand why nobody else has the same problem.


ayushin:
disregard

PR #210, opened 2020-11-03
[MIG] attachment modules: Migration to 14.0

p-tombez:
The file `attachment_s3/migrations/13.0.0.0.1/post-migration.py` was automatically removed by the module migration script, but it may be still relevant => to be confirmed


leemannd (approved):

StephaneMangin (approved):

guewen:
@p-tombez did you test them in a real bucket and with different scenarios? (Create, write, migration from db/filestore attachment to object storage, check if assets and small images are kept in DB). Every version I had to fix things, months after the migration was done, misbehaving in odds ways due to small changes in the core modules. Needless to say it took much more time to fix after all the files were in the wrong place 😜

dnplkndll:
We use this module and this version is not working on our dataset. I can provide some examples if useful. 
Env is s3 on digital ocean. Worked great on the 13.0 for a 5G dms setup. I produce an error just on module install / upgrades now.

p-tombez:
@guewen Yes, I tested them and I also had to do some minor changes
@dnplkndll Can you open an issue with as much details as possible ?

guewen:
Cool @p-tombez ! Thanks 💙

dnplkndll:
@p-tombez so actually the base_attachment_object_storage seems to need the same base64 remove. I was running some data on that. The I guess I can try a PR.

PR #213, opened 2020-11-04
[FIX] attachment modules: remove base64 encoding when reading files from storage

p-tombez:
Odoo now awaits binary content instead of base64 encoded content when reading the attachments


mmequignon (approved):

StephaneMangin (approved):

ajaniszewska-dev (approved):
LG

chroi (approved):
I'm trying hard to find a flaw ;-)

PR #220, opened 2021-03-29
[IMP] session_redis: allow configuring Redis URL instead of host & port.

theoo:
Back-porting from [13] branch


yvaucher (approved):

yvaucher:
Thanks @theoo

Issue #223, opened 2021-05-11
[RFC] attachment_s3: Support for database operations


max3903:
We would like to add the support for the following database operations with the Odoo DB manager:

* creation: When the module is installed, check the environment variables, set the config parameters, create the bucket and run force_storage().
* deletion: When the database is dropped, delete the bucket.
* duplication: When the database is duplicated, duplicate the bucket and update the store_fname in the ir_attachment table:

```
UPDATE ir_attachment AS t 
SET store_fname = s.store_fname 
FROM (
    SELECT id,REPLACE(store_fname, 'old', 'new') AS store_fname 
    FROM ir_attachment 
    WHERE db_datas is NULL)
AS s(id,store_fname)
WHERE t.id = s.id;
```

In version 12 and 14.

max3903:
@hiren-serpentcs

jjscarafia:
Hi @max3903 
I've similar needs, please check https://github.com/camptocamp/odoo-cloud-platform/issues/233 and https://github.com/camptocamp/odoo-cloud-platform/issues/234 In case you share those ideas.
Regarding your suggestions (that would be complementary to the ones I've share), this is my POV:

- creation: When the module is installed, check the environment variables, set the config parameters and run force_storage(). --> I would remove the creation of the bucket so that the environment keys don't require rights to create buckets, may be it could be optional or check if it exists or not and both use cases would be covered (in my use case I would create the bucket before calling the create database)
- deletion: When the database is dropped, delete the bucket. --> same as before, I prefer not, but if it's optional, no problem
- duplication: --> check my prs, In my mind nothing would be needed when duplicating, the idea would be to use same bucket and existing objects

If you like to share some thoughts about the big picture, we can schedule a call

BR

max3903:
@jjscarafia The big picture I have is to replicate what Odoo does by default on the filesystem, so it is transparent to admin users. They can keep working the way they do without any limitations/risks as to where the filestore is.

PR #225, opened 2021-05-20
[ADD] attachment_azure

### LINKED ISSUE #222: [ADD] attachment_azure
Hello,

We have a customer who will run Odoo in Azure using Azure Blob Storage (https://azure.microsoft.com/en-us/services/storage/blobs/) for the Odoo filestore.

So we are planning to create a new module `attachment_azure`, similar to `attachment_s3`.

Do you mind hosting the module in this repo?

Thank you.

max3903:
Closes #222 

Still missing #223 

CC @hiren-serpentcs

### DIFF
--- a/attachment_azure/models/ir_attachment.py
+++ b/attachment_azure/models/ir_attachment.py
@@ -1,0 +1,165 @@
+# Copyright 2016-2019 Camptocamp SA
+# Copyright 2021 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+import io
+import logging
+import os
+from datetime import datetime, timedelta
+
+from odoo import _, api, exceptions, models
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from azure.storage.blob import (
+        BlobServiceClient,
+        generate_account_sas,
+        ResourceTypes,
+        AccountSasPermissions,
+    )
+    from azure.core.exceptions import ResourceExistsError, HttpResponseError
+except ImportError:
+    _logger.debug("Cannot 'import azure-storage-blob'.")
+
+
+class IrAttachment(models.Model):
+    _inherit = "ir.attachment"
+
+    def _get_stores(self):
+        l = ["azure"]
+        l += super(IrAttachment, self)._get_stores()
+        return l
+
+    @api.model
+    def _get_blob_service_client(self):
+        """Connect to Azure and return the blob service client
+
+        The following environment variables must be set:
+        * ``AZURE_STORAGE_CONNECTION_STRING``
+        or
+        * ``AZURE_STORAGE_ACCOUNT_NAME``
+        * ``AZURE_STORAGE_ACCOUNT_URL``
+        * ``AZURE_STORAGE_ACCOUNT_KEY``
+
+        """
+        connect_str = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+        account_name = os.environ.get("AZURE_STORAGE_ACCOUNT_NAME")
+        account_url = os.environ.get("AZURE_STORAGE_ACCOUNT_URL")
+        account_key = os.environ.get("AZURE_STORAGE_ACCOUNT_KEY")
+        if not (connect_str or (account_name and account_url and account_key)):
+            msg = _(
+                "If you want to read from the Azure container, you must provide the "
+                "following environment variables:\n"
+                "* AZURE_STORAGE_CONNECTION_STRING\n"
+                "or\n"
+                "* AZURE_STORAGE_ACCOUNT_NAME\n"
+                "* AZURE_STORAGE_ACCOUNT_URL\n"
+                "* AZURE_STORAGE_ACCOUNT_KEY\n"
+            )
+            raise exceptions.UserError(msg)
+        blob_service_client = None
+        if connect_str:
+            try:
+                blob_service_client = BlobServiceClient.from_connection_string(
+                    connect_str
+                )
+            except HttpResponseError as error:
+                _logger.exception(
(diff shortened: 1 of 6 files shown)


yvaucher:
Nice! I'll check that next week.

p-tombez:
Looks promising. Would you consider adding the same mechanic for the storage name as in #224 ?

max3903:
@p-tombez It is already taken care of here: https://github.com/camptocamp/odoo-cloud-platform/pull/225/files#diff-2f8200ae96993037ee309fce2e5b6eaae1bcfabc8613bc549dfe2b46c29c6285R96

p-tombez:
@max3903 I don't think it's the same. Here, we don't have any control on the name instead of having an environment variable like in #224 where you can actually choose a name and add a placeholder in it for the DB name

max3903:
@p-tombez Azure does not have the concept of a bucket. You create a storage account and a shared access signature and you get:
- a connection string
- a signature
- a sas url

With those credentials, we create a container for each database. So the storage account is multi tenant.

Additional argument for this, when Odoo uses the filesystem, sysadmin don't have the control on the name of the directory.

max3903:
@p-tombez 

```
$ cat ~/.config/rclone/rclone.conf
[remote]
type = azureblob
sas_url = https://<storageaccountname>.blob.core.*****.net/?<shared_access_signature>
$ rclone lsd remote:
          -1 2021-05-26 21:58:42        -1 backup
          -1 2021-05-22 17:20:16        -1 production-master
          -1 2021-05-27 02:13:57        -1 qa-backup
          -1 2021-05-22 01:52:30        -1 qa-test
          -1 2021-05-27 21:35:13        -1 test-20210527
          -1 2021-05-27 21:09:46        -1 test-backup
```

attachment_azure/models/ir_attachment.py
```diff
@@ -0,0 +1,165 @@
...
+        * ``AZURE_STORAGE_ACCOUNT_NAME``
+        * ``AZURE_STORAGE_ACCOUNT_URL``
+        * ``AZURE_STORAGE_ACCOUNT_KEY``
+
+        """
+        connect_str = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+        account_name = os.environ.get("AZURE_STORAGE_ACCOUNT_NAME")
+        account_url = os.environ.get("AZURE_STORAGE_ACCOUNT_URL")
+        account_key = os.environ.get("AZURE_STORAGE_ACCOUNT_KEY")
+        if not (connect_str or account_name or account_url or account_key):
```
yvaucher: Looking at the error message, it seems we want to test either:

`connect_str` is provided.

OR

All three other variable are provided.

Current test will not throw the exception if for instance only `AZURE_STORAGE_ACCOUNT_URL` is provided.

```suggestion
        if not (connect_str or (account_name and account_url and account_key)):
```

attachment_azure/models/ir_attachment.py
```diff
@@ -0,0 +1,165 @@
...
+                _logger.exception(
+                    "Error during the connection to Azure container using the Shared "
+                    "Access Signature (SAS)"
+                )
+                raise exceptions.UserError(str(error))
+        return blob_service_client
+
+    @api.model
+    def _get_azure_container(self):
+        running_env = os.environ.get("RUNNING_ENV", "dev")
```
yvaucher, line 95: Might be worth documenting the `RUNNING_ENV` variable in the README. Where it comes from and what is the impact on the container name.

attachment_azure/models/ir_attachment.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @api.model
+    def _store_file_read(self, fname, bin_size=False):
+        if fname.startswith("azure://"):
+            container_client = self._get_azure_container()
+            key = fname.replace("azure://", "", 1).lower()
+            try:
+                blob_client = container_client.get_blob_client(key)
+                read = blob_client.download_blob().readall()
+            except exceptions.UserError:
```
yvaucher: This catch seems suspicious, `container_client` and `blob_client` are azure lib objects, I strongly doubt it raises Odoo like exceptions.

attachment_azure/models/ir_attachment.py
```diff
@@ -0,0 +1,165 @@
...
+            with io.BytesIO() as file:
+                blob_client = container_client.get_blob_client(key.lower())
+                file.write(bin_data)
+                file.seek(0)
+                filename = "azure://%s" % (key)
+                try:
+                    blob_client.upload_blob(file, blob_type="BlockBlob")
+                except ResourceExistsError:
+                    pass
+                except exceptions.UserError as error:
```
yvaucher: Is this `except` at the right level? Also seems suspicious, as I doubt `upload_blob` will ever raise such an exception.

attachment_azure/models/ir_attachment.py
```diff
@@ -0,0 +1,165 @@
...
+        if fname.startswith("azure://"):
+            container_client = self._get_azure_container()
+            key = fname.replace("azure://", "", 1).lower()
+            # delete the file only if it is on the current configured container
+            # otherwise, we might delete files used on a different environment
+            try:
+                blob_client = container_client.get_blob_client(key)
+                blob_client.delete_blob()
+                _logger.info("File %s deleted on the object storage" % (fname))
+            except exceptions.UserError:
```
yvaucher: same here

yvaucher (changes requested):
Thanks for your work on this module

Here are few remarks on Exception handling and for the parameters.

max3903:
@yvaucher Thanks for the review. I believe all the comments have been taken care of.

yvaucher (approved):

PR #229, opened 2021-05-28
[FIX] Bucket name formatting

mlaitinen:
The formatted value was not really assigned to `bucket_name`, so the previous commit didn't really have any effect.


yvaucher:
Seems my last change was only backward compatible :sweat_smile: 

Thanks for the fix, I'll apply your patch on all versions.

PR #237, opened 2021-07-15
[14.0] Database Operations for attachment_s3

### LINKED ISSUE #223: [RFC] attachment_s3: Support for database operations

hiren-serpentcs:
Fixes #223

### DIFF
--- a/attachment_s3/controllers/main.py
+++ b/attachment_s3/controllers/main.py
@@ -12,7 +12,7 @@ _logger = logging.getLogger(__name__)
 class Database(Database):
     @http.route()
     def drop(self, master_pwd, name):
-        delete_on_drop = os.environ.get("AWS_DELETE_ON_DBDROP")
+        delete_on_drop = os.environ.get("AWS_DELETE_ON_DBDROP", False)
         if not delete_on_drop:
             return super(Database, self).drop(master_pwd, name)
         try:
--- a/attachment_s3/controllers/main.py
+++ b/attachment_s3/controllers/main.py
@@ -1,5 +1,5 @@
 import logging
-import boto3
+import os
 
 from odoo.addons.web.controllers.main import Database
 from odoo import http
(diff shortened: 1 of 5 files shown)


attachment_s3/controllers/main.py
```diff
@@ -0,0 +1,56 @@
...
+            """ (bucket_to_copy, new_bucket_name,))
+            request.env.cr.execute(sql)
+            for key in s3_resource.list_objects(Bucket=bucket_to_copy)["Contents"]:
+                files = key["Key"]
+                copy_source = {"Bucket": bucket_to_copy, "Key": files}
+                s3_resource.meta.client.copy(copy_source, new_bucket_name, files)
+
+            return response
+        except exceptions.UserError:
+            _logger.exception("Error reading attachments from object storage.")
```
max3903:
```suggestion
            _logger.exception("Error writing attachments to object storage.")
```

max3903:
Superseeded by #245. Can be closed.

nbessi (commented):
That is more the role of an operator. That tend to goes against cloud philosophy. 
To allows more flexiblity and respect of eventually security policies I would prefer to have that feathure as a spearated addons. Also it would avoid to have a too bulky addon

PR #239, opened 2021-07-19
[13.0][ADD] monitoring_prometheus

p-tombez:
Add module to collect metrics and allow a Prometheus like server to fetch them.
Added metrics are : 
* Longpolling request count
* Request completion time

### DIFF
--- a/monitoring_prometheus/__manifest__.py
+++ b/monitoring_prometheus/__manifest__.py
@@ -16,7 +16,7 @@
     "website": "http://www.camptocamp.com",
     "data": [],
     "external_dependencies": {
-        "python": ["prometheus-client"],
+        "python": ["prometheus_client"],
     },
     "installable": True,
 }
--- a/monitoring_prometheus/__manifest__.py
+++ b/monitoring_prometheus/__manifest__.py
@@ -1,0 +1,22 @@
+# Copyright 2016-2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+
+{
+    "name": "Monitoring: Prometheus Metrics",
+    "version": "13.0.1.0.0",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "category",
+    "depends": [
+        "base",
+        "web",
+        "server_environment",
+    ],
+    "website": "http://www.camptocamp.com",
+    "data": [],
+    "external_dependencies": {
+        "python": ["prometheus-client"],
+    },
+    "installable": True,
+}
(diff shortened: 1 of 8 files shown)


leemannd (approved):
I don't know much, but this seems legit

sebalix (commented):
I don't know at all Prometheus, does it makes sense to add the number of SQL queries and query time per HTTP request here? https://github.com/odoo/odoo/blob/7cb45ec0c508b24d7966cdfea0d5d4cd9f5e203f/odoo/http.py#L1416-L1417

Maybe you need to add `prometheus-client` in [`requirements.txt`](https://github.com/camptocamp/odoo-cloud-platform/blob/13.0/requirements.txt) to make Travis happy

requirements.txt
```diff
@@ -7,3 +7,4 @@ python-keystoneclient==3.22.0
 keystoneauth1==3.14.0
 # error with 5.x (ConstructorError: could not determine a constructor for the tag '!record')
 PyYAML==4.2b4
+prometheus_client==0.11.0
```
sebalix, line 10: TIL that `pip` or `PyPI` accept `_` instead of a `-` in package names :)

p-tombez:
@sebalix The query time per HTTP request is already in the metrics. I'm not sure if the number of SQL queries is relevant as the global query time is already influenced by that.
It could be a great addition for code audit but I'm not certain it has its place as an application metric.

sebalix (approved):

sebalix:
@p-tombez ok, thanks!

monitoring_prometheus/controllers/prometheus_metrics.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2016-2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo.http import Controller, route
+from prometheus_client import generate_latest
+
+
+class PrometheusController(Controller):
+    @route('/metrics', auth='public')
```
TDu, line 9: I have two questions on the route.
* The endpoint naming could be more specific something like '/metrics/prometheus' ?
* Would it not be best practice to have some security on the end point if not now at least a todo remark somewhere ?

monitoring_prometheus/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2016-2019 Camptocamp SA
```
TDu, line 1:
```suggestion
# Copyright 2016-2021 Camptocamp SA
```

monitoring_prometheus/controllers/prometheus_metrics.py
p-tombez: 1. By default, a prometheus server will check `/metrics`
2. Firstly, the data available at this endpoint is really low-level, it's only numbers. The only risk there is a DDOS attack on the endpoint but this is already mitigated at the load balancer level. Secondly, adding security here means that we need to develop a custom collector on Prometheus side to be able to authenticate on Odoo before fetching the data and then have a specific configuration for each Odoo instance instead of having a generic one

TDu: Thanks for your answer

TDu (approved):

yvaucher (approved):

yvaucher:
squashed merged

PR #244, opened 2021-07-29
[13.0][IMP] Make cloud_platform fully abstract + update related modules

p-tombez:
Make `cloud_platform` module fully abstract and update `cloud_platform_exoscale` and `cloud_platform_ovh` accordingly.
This is to simplify the addition of other platform providers

### DIFF
--- a/cloud_platform/songs.py
+++ b/cloud_platform/songs.py
@@ -1,7 +1,3 @@
 
-def install_exoscale(ctx):
-    ctx.env['cloud.platform'].install('exoscale')
-
-
-def install_ovh(ctx):
-    ctx.env['cloud.platform'].install('ovh')
+def install(ctx):
+    ctx.env['cloud.platform'].install()
--- a/cloud_platform_exoscale/models/cloud_platform.py
+++ b/cloud_platform_exoscale/models/cloud_platform.py
@@ -1,23 +1,25 @@
 # Copyright 2016-2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
-import logging
+import re
+import os
 
 from odoo import models, api
+from odoo.addons.cloud_platform.models.cloud_platform import FilestoreKind
+from odoo.addons.cloud_platform.models.cloud_platform import PlatformConfig
 
-_logger = logging.getLogger(__name__)
 
-try:
-    from odoo.addons.cloud_platform.models.cloud_platform import FilestoreKind
-    from odoo.addons.cloud_platform.models.cloud_platform import PlatformConfig
-except ImportError:
-    FilestoreKind = None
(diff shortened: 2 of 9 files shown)


cloud_platform/songs.py
```diff
@@ -1,7 +1,3 @@
 
-def install_exoscale(ctx):
-    ctx.env['cloud.platform'].install('exoscale')
-
-
-def install_ovh(ctx):
-    ctx.env['cloud.platform'].install('ovh')
+def install(ctx):
+    ctx.env['cloud.platform'].install()
```
leemannd, line 3: This means that we will need to add manually the installation of a related platform module?

leemannd (approved):
Quick review.
Seems legit, but I'm no expert

cloud_platform_exoscale/models/cloud_platform.py
```diff
@@ -27,15 +29,81 @@ def _platform_kinds(self):
...
 
     @api.model
-    def install_exoscale(self):
-        self.install('exoscale')
+    def _check_filestore(self, environment_name):
+        params = self.env['ir.config_parameter'].sudo()
+        use_s3 = (params.get_param('ir_attachment.location') ==
+                  S3_STORE_KIND.name)
+        if environment_name in ('prod', 'integration'):
+            # Labs instances use swift or s3 by default, but we don't want
```
leemannd:
```suggestion
            # Labs instances use s3 by default, but we don't want
```
As we are now in exoscale, by default it will use only s3 correct?

cloud_platform_ovh/models/cloud_platform.py
```diff
@@ -27,15 +30,85 @@ def _platform_kinds(self):
...
 
     @api.model
-    def install_ovh(self):
-        self.install('ovh')
+    def _check_filestore(self, environment_name):
+        params = self.env['ir.config_parameter'].sudo()
+        use_swift = (params.get_param('ir_attachment.location') ==
+                     SWIFT_STORE_KIND.name)
+        if environment_name in ('prod', 'integration'):
+            # Labs instances use swift or s3 by default, but we don't want
```
leemannd:
```suggestion
            # Labs instances use swift by default, but we don't want
```
As we are into the OVH module, we will use only swift, correct?

leemannd:
Maybe we should update the readme -> https://github.com/camptocamp/odoo-cloud-platform/tree/13.0/cloud_platform  (and in the other modules also).
As we have moved the checks to the other modules.

leemannd (approved):

leemannd (approved):

leemannd (approved):

yvaucher (approved):

yvaucher (approved):

yvaucher:
squashed and merged

PR #245, opened 2021-07-30
[IMP] attachment_s3: Support for db operations

### LINKED ISSUE #223: [RFC] attachment_s3: Support for database operations

max3903:
Closes #223

### DIFF
--- a/attachment_s3/README.rst
+++ b/attachment_s3/README.rst
@@ -11,14 +11,37 @@ Activate S3 storage:
 
 * Create or set the system parameter with the key ``ir_attachment.location``
   and the value in the form ``s3``.
+* Configure accesses with the following environment variables:
 
-Configure accesses with environment variables:
+.. list-table::
+   :header-rows: 1
 
-* ``AWS_HOST`` (not required if using AWS services)
-* ``AWS_REGION`` (required if using AWS services)
-* ``AWS_ACCESS_KEY_ID``
-* ``AWS_SECRET_ACCESS_KEY``
-* ``AWS_BUCKETNAME`` (optional {db} placeholder)
+   * - Name
+     - Description
+     - Values
+   * - ``AWS_HOST``
+     - Endpoint.
+     - Not required if using AWS (defaults to s3.amazonaws.com).
+   * - ``AWS_REGION``
+     - Region
+     - Required if using AWS.
+   * - ``AWS_ACCESS_KEY_ID``
+     - Access Key ID
+     -
+   * - ``AWS_SECRET_ACCESS_KEY``
+     - Secret Access Key
+     -
+   * - ``AWS_BUCKETNAME``
+     - Name of the bucket (AWS) or space (Digital Ocean)
+     - Optional {db} placeholder
+   * - ``AWS_DUPLICATE``
+     - If set, the bucket and all its objects will be copied when the database is
+       duplicated. The paths in the database will be updated to use the copy.
+     - True
+   * - ``AWS_EMPTY_ON_DBDROP``
+     - If set, all the objects in the bucket will be deleted when the database is
+       dropped.
+     - True
 
 Read-only mode:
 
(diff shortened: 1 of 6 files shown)


max3903:
This has been tested successfully with AWS and Digital Ocean.

max3903:
ping @jevinD

max3903:
Ping @jjscarafia

attachment_s3/controllers/main.py
```diff
@@ -0,0 +1,73 @@
...
+                bucket = request.env["ir.attachment"]._get_s3_bucket(bucket)
+                bucket.objects.all().delete()
+            except exceptions.UserError:
+                _logger.exception("Error deleting attachments in object storage.")
+        return res
+
+    @http.route()
+    def duplicate(self, master_pwd, name, new_name):
+        res = super().duplicate(master_pwd, name, new_name)
+        try:
```
jjscarafia, line 34: I would make it optional. May be with another environment variable?

attachment_s3/README.rst
```diff
@@ -12,13 +12,30 @@ Activate S3 storage:
 * Create or set the system parameter with the key ``ir_attachment.location``
   and the value in the form ``s3``.
 
-Configure accesses with environment variables:
```
jjscarafia, line 15: The new table is nice, but there is no reference telling that this are environment variables to be configured. I would keep something similar to this row

attachment_s3/controllers/main.py
jjscarafia: I would make this new duplication behaviour optional. May be with another environment variable?

attachment_s3/__init__.py
```diff
@@ -1,2 +1,18 @@
...
+import os
 from . import models
+from . import controllers
+
+from odoo import api, SUPERUSER_ID
+
+
+def _post_init_hook(cr, registry):
+    # create the S3 bucket after module installation
+    if os.environ.get("AWS_BUCKETNAME", False):
```
jjscarafia, line 12: regarding this post init and also de "restore" we're working in a solution that may be could help here.
The problem we face is that if you call "force_storage" for big databases, it will take a lot of time.
So we're doing a simple modification and adding a cron (Disable by default) that you can enable and then it starts to move the filestore in batch to s3.
With that optional functionality, this bot use cases would be solved.

Have you try to upload big databases? It works ok with normal odoo timeout limits?

jjscarafia (changes requested):
Hi @max3903 
Thanks for pinging.
For now we would not use any of this improvements as we would work similar to camptocamp where many databases re-use same attachments.

Still, I  have made a review and I leave some comments. IMHO all this functionalities added should optional.

We've a WIP here with the changes we're working on https://github.com/camptocamp/odoo-cloud-platform/pull/248 (we still need to split the PRs and make some other improvements)

The pr with the cron for the "force_storage" will be done on this week

attachment_s3/__init__.py
max3903: @jjscarafia We haven't tested that scenario yet.

attachment_s3/controllers/main.py
max3903: @jjscarafia ok

vrenaville:
For my point of view, playing with access policy , prevent the need to have this kind of duplication.

PR #249, opened 2021-08-06
[13.0][ADD] cloud_platform_azure

p-tombez:


cloud_platform_azure/models/cloud_platform.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @api.model
+    def _filestore_kinds(self):
+        kinds = super(CloudPlatform, self)._filestore_kinds()
+        kinds["azure"] = AZURE_STORE_KIND
+        return kinds
+
+    @api.model
+    def _platform_kinds(self):
+        kinds = super(CloudPlatform, self)._platform_kinds()
```
i-vyshnevska, line 26:
```suggestion
        kinds = super()._platform_kinds()
```

cloud_platform_azure/models/cloud_platform.py
```diff
@@ -0,0 +1,124 @@
...
+    @api.model
+    def _check_filestore(self, environment_name):
+        params = self.env["ir.config_parameter"].sudo()
+        use_azure = (params.get_param("ir_attachment.location") ==
+                     AZURE_STORE_KIND.name)
+        if environment_name in ("prod", "integration"):
+            # Labs instances use azure by default, but we don't want
+            # to enforce it in case we want to test something with a different
+            # storage. At your own risks!
+            assert use_azure, (
```
i-vyshnevska, line 51: why not logger?

cloud_platform_azure/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+        "cloud_platform",
+        "attachment_azure",
+        "monitoring_prometheus",
+    ],
+    "excludes": [
+        "cloud_platform_ovh",
+        "cloud_platform_exoscale",
+    ],
+    "website": "https://www.camptocamp.com",
+    "data": [],
```
i-vyshnevska, line 22:
```suggestion
```

cloud_platform_azure/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "version": "13.0.1.0.0",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Extra Tools",
+    "depends": [
+        "cloud_platform",
+        "attachment_azure",
+        "monitoring_prometheus",
+    ],
+    "excludes": [
```
i-vyshnevska, line 17: just wondering does it has any impact or just for information

i-vyshnevska (approved):
generally LG, some small questions

cloud_platform_azure/models/cloud_platform.py
p-tombez: It will be backported to all Odoo versions, also with python 2.7. I find it easier to have that so it's consistent between all versions even if with python 3 it's not required anymore

cloud_platform_azure/models/cloud_platform.py
p-tombez: Because we want to stop Odoo from starting if the check fails

cloud_platform_azure/__manifest__.py
p-tombez: It will show an error in the logs and prevent the installation if you try to install the module when any of those is already installed

ajaniszewska-dev (approved):

yvaucher (approved):
LGTM

PR #250, opened 2021-08-09
[12.0] Add pre-commit-hook

leemannd:
I saw a PR where travis was failing because it wasn'tcompliant with flake8 (and it was legit).

In this PR I'm adding the same pre-commit-hook that we have on our base projects.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,14 @@
+[flake8]
+max-line-length = 79
+max-complexity = 16
+# B = bugbear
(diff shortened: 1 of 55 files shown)


p-tombez (commented):
This project is based on the OCA template, so if you need to run pre-commit locally, you're supposed to use the standard OCA maintainer tools workflow.
IMO, it's not a good thing to have 2 different pre-commit configurations for the same project

PR #251, opened 2021-08-10
monitoring_status: do not save session at all

sbidoul:


sbidoul:
Unless, of course, the status endpoint also needs to check the session store.

max3903:
If users cannot login, I would prefer the check to fail...

sbidoul:
yeah, it's probably better to keep it in this context where sessions are in redis

PR #257, opened 2021-08-23
[WIP][13.0][IMP] base_attachment_object_storage: execute force_storage function with batch limit

lisandrogallo:
Added a scheduled action for executing the 'force_storage' function to upload objects to S3 using a batch limit


vrenaville:
Hello,
Do you really need to call it multiple time. At camptocamp we use it one time with an odoo shell command, when we migrate a project from a db storage to an external storage.

PR #259, opened 2021-09-02
[13.0][FIX] attachment_azure

nilshamerlinck:
Hi, in 13.0, `_file_read()` is expected to return base64 encoded data.


yvaucher:
@p-tombez you might want to check as you did some tests with azure. Not sure you used v13 though.

p-tombez (approved):
@yvaucher I only tested the connection, I couldn't test the file operations back then as there was permissions issues.
So, LGTM

PR #262, opened 2021-09-28
[12.0][IMP] Add S3 Bucket Registry to store clients.

yankinmax:
Each time _store_file functions are called a new S3 Bucket object was created.
Instead of creating a new one, we store the client in a Bucket Registry
to avoid checking if the Bucket exists every time.

### DIFF
--- a/attachment_s3/models/ir_attachment.py
+++ b/attachment_s3/models/ir_attachment.py
@@ -23,6 +23,18 @@ except ImportError:
     _logger.debug("Cannot 'import boto3'.")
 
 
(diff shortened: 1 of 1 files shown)


grindtildeath (approved):
:+1:

leemannd (approved):
Seems legit, I'm not the most experienced one on this topic

yvaucher (approved):

jjscarafia:
Just in case it help, I've created the FP to v13 of this https://github.com/camptocamp/odoo-cloud-platform/pull/321

Thanks!

PR #266, opened 2021-10-18
[13.0][IMP] attachment_s3: Add DB operations

max3903:


nbessi (changes requested):
Same remark as previous PR that should be in a separated addon

PR #277, opened 2021-11-03
[15.0] Forward port of cloud_platform abstraction

yvaucher:
Forwardport of #244

### DIFF
--- a/cloud_platform_ovh/models/cloud_platform.py
+++ b/cloud_platform_ovh/models/cloud_platform.py
@@ -10,7 +10,7 @@ from odoo.addons.cloud_platform.models.cloud_platform import FilestoreKind
 from odoo.addons.cloud_platform.models.cloud_platform import PlatformConfig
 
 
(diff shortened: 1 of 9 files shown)


cloud_platform_ovh/models/cloud_platform.py
```diff
@@ -1,24 +1,27 @@
...
 
-try:
-    from odoo.addons.cloud_platform.models.cloud_platform import FilestoreKind
-    from odoo.addons.cloud_platform.models.cloud_platform import PlatformConfig
-except ImportError:
-    FilestoreKind = None
-    PlatformConfig = None
-    _logger.debug("Cannot 'import from cloud_platform'")
+
+SWIFT_STORE_KIND = FilestoreKind('s3', 'remote')
```
gurneyalex: I have issues making the abstraction version on OVH and I think this is because this line should read
```suggestion
SWIFT_STORE_KIND = FilestoreKind('swift', 'remote')
```

PR #280, opened 2021-11-05
14.0 fix ovh storage

gurneyalex:


jcoux (approved):
Thanks :+1:

leemannd (commented):
Seems legit. Will need to be ported to other versions

yvaucher (approved):

PR #284, opened 2021-11-09
[IMP] attachment_s3: dont update already existis objects

jjscarafia:


nbessi (changes requested):
The test should be based on a hash equality instead of lennth no ? Else it might induce unexpected behavior by the user. Even if I agree 99% of the case odoo will not replace in place attachement

jjscarafia:
hi @nbessi 
Thanks for checking this.
Agree that content_length should be improved, we just check something exists.
The use case here is not to edit an attachment, the use case is to upload the same file two times (for eg. on tho different records).
For odoo they are two attachments but the hash is only one, and when pusing to s3 we're writing the previous file with this issues:
a) we have a policy on the s3 to protect files, so they can't be written
b) writing same content on the s3 object is not actually needed.

Thanks!

PR #307, opened 2021-11-12
[13.0][IMP] attachment_azure: Write and read storage name along filename for extended usability

p-tombez:
Store the name of the storage container along the filename in order to be able to use multiple containers for one database


yvaucher (approved):
Change LGTM

Can you add a line about it in README ?

PR #322, opened 2022-01-04
fix: do not try to create container every time

vrenaville:
* Purpose
In some case you want to read an other container that you have access only on readonly
ex: read production container datas from your validation instance.
Without this patch, odoo will check that the container exist by trying to create it, and it will raise an azure error, unauthorized.
We simply list available container, and check if the container that we need to access is in the list, if not we try to create it

### DIFF
--- a/attachment_azure/models/ir_attachment.py
+++ b/attachment_azure/models/ir_attachment.py
@@ -134,14 +134,13 @@ class IrAttachment(models.Model):
             container_name = self._get_container_name()
         blob_service_client = self._get_blob_service_client()
         container_client = blob_service_client.get_container_client(container_name)
-        try:
-            # Create the container
-            container_client.create_container()
-        except ResourceExistsError:
-            pass
-        except HttpResponseError as error:
-            _logger.exception("Error during the creation of the Azure container")
-            raise exceptions.UserError(str(error))
+        if not container_client.exists():
+            try:
+                # Create the container
+                container_client.create_container()
+            except HttpResponseError as error:
+                _logger.exception("Error during the creation of the Azure container")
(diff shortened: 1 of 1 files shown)


nbessi (approved):

attachment_azure/models/ir_attachment.py
```diff
@@ -134,14 +134,13 @@ def _get_azure_container(self, container_name=None):
...
         container_client = blob_service_client.get_container_client(container_name)
-        try:
-            # Create the container
-            container_client.create_container()
-        except ResourceExistsError:
-            pass
-        except HttpResponseError as error:
-            _logger.exception("Error during the creation of the Azure container")
-            raise exceptions.UserError(str(error))
+        if not any(x['name'] == container_name for x in blob_service_client.list_containers()):
```
yvaucher: `list_containers` has a parameter `name_starts_with`, wouldn't it be more effective to ask only for matching container names?

 https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.blobserviceclient?view=azure-python#list-containers-name-starts-with-none--include-metadata-false----kwargs-

vrenaville: @yvaucher I have seen this, it would be great in our case, but for other person it may be an other use case

yvaucher: Have you tried using the method `containter_client.exists()`?
https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.containerclient?view=azure-python#exists---kwargs-

yvaucher (approved):

PR #344, opened 2022-02-08
BSQOQ-647: allow instance profiles authentication

mfournier:
By dropping the requirement to have `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` envvars set, we can let the more secure IAM instance-profile take over the authentication.

This can be used immediatly on the rancher platform, and NB that this will become a requirement on the kubernetes platform in a couple of weeks.

### DIFF
--- a/attachment_s3/models/ir_attachment.py
+++ b/attachment_s3/models/ir_attachment.py
@@ -58,10 +58,12 @@ class IrAttachment(models.Model):
         # replaces {db} by the database name to handle multi-tenancy
         bucket_name = bucket_name.format(db=self.env.cr.dbname)
         params = {
-            'aws_access_key_id': access_key,
-            'aws_secret_access_key': secret_key,
             'bucket_name': bucket_name,
         }
+        if access_key:
+            params['aws_access_key_id'] = access_key
+        if secret_key:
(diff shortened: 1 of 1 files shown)


p-tombez (changes requested):
Globally, looks OK, but you can't let the parameters unchecked.
So, I propose to add a new boolean envvar `AWS_USE_PROFILE`, then update the test in line 91 to something like
```
if not (
    bucket_name and
    (params["aws_access_key_id"] and
     params["aws_secret_access_key"] or
     params["aws_use_profile"])
):
```

Please also update : 
* Comment on line 78 to add the new envvar
* Error message on line 96
* Module documentation

nbessi:
We will do a new proper PR @mfournier .

mfournier:
Thanks @p-tombez & @nbessi for the update !

I'm fine with @p-tombez's suggestion, and of course happy to let @nbessi take over this PR :wink: 

Just one detail: in AWS terminology a "profile" and an "instance-profile" are different things. I suggest naming the feature/variables/etc `aws_use_instance_profile` to avoid confusion.

Thanks guys :muscle:

PR #349, opened 2022-03-03
fix: fix subdirectory issue

vrenaville:


StephaneMangin:
Does the script that migrate from S3 to azure takes this in consideration too ?

PR #356, opened 2022-03-18
[12.0] use instance profile for S3 storage

vrenaville:
* Purpose
It replace this PR:
https://github.com/camptocamp/odoo-cloud-platform/pull/344
I had instance_profile env

### DIFF
--- a/attachment_s3/models/ir_attachment.py
+++ b/attachment_s3/models/ir_attachment.py
@@ -54,6 +54,7 @@ class IrAttachment(models.Model):
         region_name = os.environ.get('AWS_REGION')
         access_key = os.environ.get('AWS_ACCESS_KEY_ID')
         secret_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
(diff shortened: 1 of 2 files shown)


p-tombez (approved):

yann-soubeyrand:
Technically speaking, we don’t want to use instance profile (the IAM role attached to the EKS node), but the IAM role attached to the Kubernetes service account.

attachment_s3/models/ir_attachment.py
```diff
@@ -79,6 +84,7 @@ def _get_s3_bucket(self, name=None):
         * ``AWS_ACCESS_KEY_ID``
         * ``AWS_SECRET_ACCESS_KEY``
         * ``AWS_BUCKETNAME``
+        * ``AWS_USE_INSTANCE_PROFILE``
```
yann-soubeyrand: Shouldn’t it be `AWS_USE_IRSA`?

PR #364, opened 2022-04-13
Object Storage - disable mode

StephaneMangin:
To prevent issues while migrating locally in integration/production mode, here is a proposal to add an environment variable to avoid storage access during migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,3 +162,8 @@ environment. It will refuse to start if anything is badly configured.
 
 The checks can be bypassed with the environment variable
 `ODOO_CLOUD_PLATFORM_UNSAFE` set to `1`.
+
+### Attachment storage inactivation 
+
+To prevent object storage to be accessed while failing for any kind of reason
+set this environment variable `ATTACHMENT_STORAGE_INACTIVE` set to `1`.
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ environment. It will refuse to start if anything is badly configured.
 The checks can be bypassed with the environment variable
 `ODOO_CLOUD_PLATFORM_UNSAFE` set to `1`.
 
-### Attachment storage inactivation 
+### Attachment storage disability 
 
 To prevent object storage to be accessed while failing for any kind of reason
-set this environment variable `ATTACHMENT_STORAGE_INACTIVE` set to `1`.
+set this environment variable `DISABLE_ATTACHMENT_STORAGE` set to `1`.
--- a/base_attachment_object_storage/README.rst
+++ b/base_attachment_object_storage/README.rst
@@ -1,7 +1,7 @@
 Base class for attachments on external object store
 ===================================================
 
-This is a base addon that regroup common code used by addons targeting specific object store 
+This is a base addon that regroup common code used by addons targeting specific object store
 
 Configuration
 -------------
@@ -38,3 +38,9 @@ Default configuration means:
   stored in database
 * application/javascript are stored in database whatever their size
 * text/css are stored in database whatever their size
+
+Inactivate attachment storage I/O
+---------------------------------
+
+Define a environment variable `ATTACHMENT_STORAGE_INACTIVE` set to `1`
+This will prevent any kind of exceptions and read/write on storage attachments.
--- a/base_attachment_object_storage/README.rst
+++ b/base_attachment_object_storage/README.rst
@@ -39,8 +39,8 @@ Default configuration means:
 * application/javascript are stored in database whatever their size
 * text/css are stored in database whatever their size
 
-Inactivate attachment storage I/O
----------------------------------
+Disable attachment storage I/O
+------------------------------
 
-Define a environment variable `ATTACHMENT_STORAGE_INACTIVE` set to `1`
+Define a environment variable `DISABLE_ATTACHMENT_STORAGE` set to `1`
 This will prevent any kind of exceptions and read/write on storage attachments.
--- a/base_attachment_object_storage/models/ir_attachment.py
+++ b/base_attachment_object_storage/models/ir_attachment.py
@@ -5,6 +5,7 @@ import inspect
 import logging
 import os
 import time
+from distutils.util import strtobool
 
 import psycopg2
 import odoo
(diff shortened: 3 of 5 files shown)


santostelmo (dismissed):

base_attachment_object_storage/README.rst
```diff
@@ -38,3 +38,10 @@ Default configuration means:
   stored in database
 * application/javascript are stored in database whatever their size
 * text/css are stored in database whatever their size
+
+Unsafe mode
+-----------
+
+Define a environment variable `ATTACHMENT_STORAGE_UNSAFE` set to `1`
+This will prevent and avoid any king of exceptions and filestore read/write
```
simahawk:
```suggestion
This will prevent any kind of exceptions and any read/write on filestore 
```

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -151,6 +156,12 @@ def _store_in_db_instead_of_object_storage(self, data, mimetype):
         ``_store_in_db_instead_of_object_storage_domain``.
 
         """
+        unsafe_mode = is_true(os.environ.get("ATTACHMENT_STORAGE_UNSAFE"))
```
simahawk: can you move this to 
```
if self._unsafe_mode_on():
    return True
```
and move the code there to be reused?

base_attachment_object_storage/README.rst
guewen: Hey, if I read "unsafe", I read it as it "activate the unsafe mode in which they ar not guarantees", which means kind of the reverse of what it is doing.
Maybe it's only me.

StephaneMangin: > Hey, if I read "unsafe", I read it as it "activate the unsafe mode in which they ar not guarantees", which means kind of the reverse of what it is doing. Maybe it's only me.

Hi @guewen
Meaning of unsafe could be a lot of interpretations in fact, but you're right. In this case "unsafe" means that you can't garrantee filestore interactions as it is almost totally deactivated. Which means you can't read or write on a storage that can't be accessible, which is not safe to use in this context because of possible data loss.

Have you a propasal for this definition then ?  INACTIVE or something ?

StephaneMangin: And it was made to follow the same logic as the unsafe mode of the cloud_platform module

StephaneMangin: Changed for "inactive" concept instead of "unsafe" !

leemannd (approved):

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -40,6 +45,18 @@ def clean_fs(files):
 class IrAttachment(models.Model):
     _inherit = 'ir.attachment'
 
+    @staticmethod
+    def is_storage_inactive(storage=None, log=True):
```
simahawk: log is never used, am I wrong?

base_attachment_object_storage/models/ir_attachment.py
```diff
@@ -18,6 +19,10 @@
 _logger = logging.getLogger(__name__)
 
 
+def is_true(strval):
+    return bool(strtobool(strval or '0'))
```
sebalix, line 23: Can we use `str2bool` provided by Odoo instead?
https://github.com/odoo/odoo/blob/14.0/odoo/tools/convert.py#L206-L207=

base_attachment_object_storage/models/ir_attachment.py
sebalix: It is

sebalix (approved):

base_attachment_object_storage/README.rst
yankinmax: @StephaneMangin thanks, checked on customers project

yankinmax (approved):

base_attachment_object_storage/README.rst
guewen: The unsafe param in the cloud platform module does not disable any functionality, it allows to start odoo even if some checks would not be valid, that's not really the same idea then.

Disable is better 👍

sebalix: I like `DISABLE`, it makes the thing clear about the intent.

santostelmo (approved):

santostelmo:
@StephaneMangin pre-commit

base_attachment_object_storage/README.rst
StephaneMangin: Done !

README.md
```diff
@@ -162,3 +162,8 @@ environment. It will refuse to start if anything is badly configured.
 
 The checks can be bypassed with the environment variable
 `ODOO_CLOUD_PLATFORM_UNSAFE` set to `1`.
+
+### Attachment storage inactivation 
+
+To prevent object storage to be accessed while failing for any kind of reason
+set this environment variable `ATTACHMENT_STORAGE_INACTIVE` set to `1`.
```
sebalix:
```suggestion
set this environment variable `DISABLE_ATTACHMENT_STORAGE` set to `1`.
```

base_fileurl_field/fields.py
```diff
@@ -61,7 +61,7 @@ def create(self, record_values):
                         storage_key = self._build_storage_key(fname)
                 if not fname:
                     storage_key = False
-                env['ir.attachment'].sudo().with_context(
+                super().with_env(env).sudo().with_context(
```
sebalix: Are you sure about this? You switched the `create` call from `ir.attachment` to the one of the field.

StephaneMangin: Hum indeed,  rollbacked !

README.md
StephaneMangin: Pushed ^^

sebalix: Maybe GitHub is tricking me, but now it's `ATTACHMENT_STORAGE_INACTIVE`

yankinmax:
@StephaneMangin I'll update v15 PR to be able to merge too)

PR #367, opened 2022-05-03
[15.0][IMP] attachment_s3: add skip_objects_deletion

jjscarafia:
Port of https://github.com/camptocamp/odoo-cloud-platform/pull/283/


nbessi (commented):
I'm not against the Idea @p-tombez what is you opinion. That might be duplicate with S3 soft delete

vrenaville:
I'm not against, it's a minimal impact, It may lead to a lot of dead object in the storage.
but this issue can be managed outside of odoo

vrenaville (approved):

PR #368, opened 2022-05-03
[15.0][IMP] attachment_s3: dont update already existis objects

jjscarafia:
Port of https://github.com/camptocamp/odoo-cloud-platform/pull/284


nbessi (changes requested):
That test seems a little to light. 
An hash comparison should be added

nbessi (changes requested):
That test seems a little to light. 
An hash comparison should be added

PR #389, opened 2022-10-25
[FIX] requirements: use cryptography==35.0.0

jjscarafia:
Odoo requires maxiumn verison of cryptography to be 35.0.0 (https://github.com/odoo/odoo/blob/15.0/requirements.txt#L3) but azure-storage-blob could take to an upgrade to latest version (38.x) With this fix we pre-intall cryptography==35.0.0 before azure-storage-blob could update it

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+cryptography==35.0.0
 azure-storage-blob==12.8.1
 azure-identity==1.6.0
 boto3==1.9.102


nilshamerlinck:
Hello @jjscarafia 

I believe it's actually `azure-identity==1.6.0` which triggers the upgrade of `cryptography` to its last version. I documented that unexpected case here: https://github.com/pypa/pip/issues/11553 (unexpected to me at least)

That said, your proposed workaround lgtm :+1:

PR #394, opened 2022-11-15
[16.0] [mig] attachment_s3

mileo:


adrienpeiffer:
I have encountered some problems with the use of the document module. I'll give you some more information soon. @vrenaville Do you plan to work on the subject?

vrenaville:
@adrienpeiffer 
We don't use S3 anymore, I'm focused on azure, I will try in next week to improve to use Stream more efficiently to load file by chunk , currently it load the file in memory in one big chunk, it's not efficient

jjscarafia:
Hi @adrienpeiffer 
We've work on a similar migration but using a different approach on PR https://github.com/camptocamp/odoo-cloud-platform/pull/401. We've some uses cases where the inheritance of _record_to_stream was not enought and the super call was triying to stream content of an s3 record (For eg. a binary field on website model or something like that).

So me work on an inheritance on an upper method on base_attachment_object_storage. 
Is working properly till now.

The load of the file is not optimal and your idea of improvement would be needed

cc @filoquin

vrenaville:
@jjscarafia 
I think @adrienpeiffer have adopt the same method that I have done for Azure storage.
My method try to avoid monkey patching, the more I can.

@gurneyalex did you have a comment on this ?

gurneyalex:
the approach re

> @jjscarafia I think @adrienpeiffer have adopt the same method that I have done for Azure storage. My method try to avoid monkey patching, the more I can.

This is on https://github.com/camptocamp/odoo-cloud-platform/pull/401

gurneyalex:
@vrenaville what is annoying is that in both approaches we have to read the attachment from memory and we are not really able to stream it. The approach in #401 is nicer in that if fixes the issue in the base module and I would go for that solution, until we have a way to stream things from the cloud provider.

jjscarafia:
I completely agree that we still need to improve and stream it.
I didn't find time yet to show a use case where this face the issue descrived [here](https://github.com/camptocamp/odoo-cloud-platform/pull/394#issuecomment-1365960046). Doing as https://github.com/camptocamp/odoo-cloud-platform/pull/401/files works till now on every use case

PR #404, opened 2023-01-25
add conditionnal mimetype to upload file in bucket s3

cyrilmanuel:
add conditionnal mimetype to upload file in bucket s3

### DIFF
--- a/attachment_s3/models/ir_attachment.py
+++ b/attachment_s3/models/ir_attachment.py
@@ -9,6 +9,7 @@ import io
 from urllib.parse import urlsplit
 
 from odoo import _, api, exceptions, models
+from odoo.addons.base.models.ir_attachment import IrAttachment as BaseIrAttachment
 from ..s3uri import S3Uri
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


yvaucher (approved):

yvaucher:
Seems the CI is failling but unrelated to that PR

```
ERROR: Could not find a version that satisfies the requirement pylint-plugin-utils==0.7 (from -r /home/travis/maintainer-quality-tools/travis/pylint_odoo_requirements.txt (line 1)) (from versions: 0.1, 0.1.1, 0.2, 0.2.1, 0.2.2, 0.2.3, 0.2.4, 0.2.6, 0.3, 0.4, 0.5, 0.6)
```

yvaucher:
We should change the unit tests to be run with github actions.

MQT now fails due to a change on pylint-odoo that requires a dependency which is too recent and dropped support for python <3.6.2.

https://github.com/OCA/oca-addons-repo-template/tree/master/.github/workflows

I'll make a quick patch.

vrenaville (approved):

gurneyalex:
@yvaucher did you make the quick patch?

attachment_s3/models/ir_attachment.py
```diff
@@ -178,9 +178,37 @@ def _store_file_read(self, fname, bin_size=False):
...
+            if value and location != 'db':
+                # save it to the filestore
+                self=self.with_context(mimetype=attach.mimetype)
+                vals['store_fname'] = self._file_write(value, vals['checksum'])
+                vals['db_datas'] = False
+
+            # take current location in filestore to possibly garbage-collect it
+            fname = attach.store_fname
+            # write as superuser, as user probably does not have write access
+            super(IrAttachment, attach.sudo()).write(vals)
```
yvaucher: :warning: With @sarsurgithub we discovered that this super is there to bypass the write of `ir.attachment` where it pops out the `file_size` and `checksum` fields as it is supposed to be written by this inverse function.


Thus here with this overwrite we loose this bypass.
:arrow_right: As a result `file_size` and `checksum` are never written.

You need to import the base ir_attachment model:

```
from odoo.addons.base.models.ir_attachment import IrAttachment as BaseIrAttachment
```
Then use the base model for the write.
```suggestion
            super(BaseIrAttachment, attach.sudo()).write(vals)
```

yvaucher (changes requested):

PR #406, opened 2023-02-14
Patch travis dependency of MQT for py3.5

yvaucher:
This is an attempt to quickly patch branch 12.0.

A better approach would be to use https://github.com/OCA/oca-addons-repo-template/tree/master/.github/workflows


yvaucher:
Closing as we will move to GH actions.

PR #421, opened 2023-04-26
[13.0] Change CI to GitHub actions

yvaucher:
Use copier template from oca/oca-addons-repo-template

Target Python3.8

Apply linting

Fix a missing call to super


test_base_fileurl_field/models/res_partner.py
```diff
@@ -21,24 +21,32 @@ class ResPartner(models.Model):
...
-        rec = self.search([("url_file_fname", "=", self.url_file_fname)])
-        if len(rec) > 1:
-            raise ValidationError(
-                _(
-                    "This file name is already used on an existing record. "
-                    "Please use another file name or delete the url_file on :\n"
-                    "Model: %s Id: %s" % (self._name, rec.id)
-                )
+        for rec in self:
+            match = self.search(
```
simahawk, line 30: would be better to use `search_count` in such cases

test_base_fileurl_field/models/res_partner.py
```diff
@@ -1,44 +1,57 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
-from odoo import models, fields, api, _
+from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
+# make pytest happy
+# in pytest context module dependencies are not loaded
+# thus FileURL is unknown
+from odoo.addons import base_fileurl_field  # noqa
```
simahawk, line 9: we should import the field directly instead of relying on a patch (as we do for base_sparse_field) :wink: 
```suggestion
from odoo.addons.base_fileurl_field.fields import FileURL
```

test_base_fileurl_field/models/res_partner.py
```diff
@@ -1,44 +1,57 @@
...
+
 
 class ResPartner(models.Model):
 
-    _inherit = 'res.partner'
+    _inherit = "res.partner"
 
     name = fields.Char()
     url_file = fields.FileURL(
```
simahawk, line 17:
```suggestion
    url_file = FileURL(
```

test_base_fileurl_field/models/res_partner.py
```diff
@@ -1,44 +1,57 @@
...
-        storage_location='s3',
-        filename='url_file_fname',
-        storage_path='partner'
+        storage_location="s3", filename="url_file_fname", storage_path="partner"
     )
     url_file_fname = fields.Char()
 
     url_image = fields.FileURL(
```
simahawk, line 22:
```suggestion
    url_image = FileURL(
```

simahawk (approved):

test_base_fileurl_field/models/res_partner.py
yvaucher: Outside of topic of this PR I'm only applying lint and making CI happy.

test_base_fileurl_field/models/res_partner.py
yvaucher: IMO we should be able to use it from `odoo.fields` from any field Type registered in `odoo.fields`.

`fields.MyFieldType` is a common usage and making exception is not great.

simahawk: IMO is just a dirty hack. In fact, it's a patch and you have no certainty that is going to be loaded at the right time. Plus, no framework out there - to my knowledge - relies on such approach and just imports the thing you need from where it stays.
Not to mention that is way more explicit :wink: 

"common usage" != "good practice" :grin:

PR #422, opened 2023-05-10
[14.0] Change CI to GitHub actions

yvaucher:
Use copier template from oca/oca-addons-repo-template

Target Python3.8

Apply linting

Fix a missing call to super


simahawk (approved):
same remarks of #421

PR #435, opened 2023-07-10
[16.0] session_redis: allow kwargs in session vacuum

Warlocklogan:
The vacuum method from the SessionStore class now takes an argument `max_lifetime`   
This change was made in the following PR: https://github.com/odoo/odoo/pull/122888  
To avoid GC errors in the latest Odoo 16 version of the session_reddis the vacuum method method can now handle any kwargs.


vincent-hatakeyama:
Tested, that solves the issue with the new argument.

PR #438, opened 2023-08-22
base_attachment_storage: _register_hook do nothing if it is disabled

leemannd:
When doing a migration and installing the module and having disabled the storage. I don't want the register hook to do anything.


leemannd:
Closing as I didn't receive any feedback

PR #468, opened 2024-08-27
[FIX] monitoring_log_requests: response_status_code

gurneyalex:
the response status code would not be properly logged -> fix the code

### DIFF
--- a/monitoring_log_requests/models/ir_http.py
+++ b/monitoring_log_requests/models/ir_http.py
@@ -6,7 +6,7 @@ import logging
 import time
 
 from odoo import models
-from odoo.http import request as http_request
+from odoo.http import Response, request as http_request
 from odoo.tools.config import config
 
 _logger = logging.getLogger("monitoring.http.requests")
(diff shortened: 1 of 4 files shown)


santostelmo (approved):

monitoring_log_requests/models/ir_http.py
```diff
@@ -62,8 +62,10 @@ def _monitoring_info(cls, request, response, begin, end):
             # response things
             "response_status_code": None,
         }
-        if hasattr(request, "status_code"):
-            info["status_code"] = response.status_code
+        if isinstance(response, Response) and hasattr(response, "status_code"):
+            info["response_status_code"] = response.status_code
+        else:
```
Ricardoalso, line 67: Can this not be misleading ? Or are we always sure that an inexistent status_code represents a 200 status code meaning that the request was successful :thinking: 
It is basically like setting `"response_status_code": 200 `in line 63

cyrilmanuel (approved):

PR #471, opened 2024-11-11
[MIG][16.0] attachment_swift

gdgellatly:
Just draft for now while testing, no changes except installable and version no in line with s3 migration.


gdgellatly:
FYI - happily running in prod, shame I forgot to add pre-commit, will redo and push

PR #472, opened 2024-12-20
[MIG] session_redis: Migration to 18.0

vib-adhoc:


requirements.txt
```diff
@@ -1,6 +0,0 @@
-# generated from manifests external_dependencies
-azure-identity
-azure-storage-blob
-prometheus_client
-python-json-logger
-redis
```
vrenaville, line 6: Why removing redis, we will need it installed no ?

vib-adhoc: Hello Vincent, you are right. That problem was because I ran pre-commit before doing the migration. I already corrected it, what do you think?

vrenaville:
@vib-adhoc I have seen an other port of session_redis here : https://github.com/camptocamp/odoo-cloud-platform/pull/478

Are you ok if we keep this port, that done some rework

vib-adhoc:
Hi @vrenaville , I have seen that the migration has already been done here: [#482](https://github.com/camptocamp/odoo-cloud-platform/pull/482). Thank you for advancing it.

PR #474, opened 2025-01-23
[FIX] Enable session_redis to work with Sentry

dnplkndll:
ref: https://github.com/avoinsystems/odoo-cloud-platform/commit/a09a8136d5e79a4f00b7a5645e87e677ada4ed1e

When the sentry module is loaded, it overrides odoo.http.Application with a middleware object that wraps the original, but does not proxy it. Thus the session_store property that session_redis monkeypatches is no longer there.

This fix intends to avoid that, without having very specific knowledge about the sentry module baked in.

The solution seems to work, but has the risk that if the Application class is fully replaced (i.e. if it is not even wrapped), then this new method may lead to this module simply not working, and the reasons for it may be difficult to debug. But the only way around that would be to explicitly check the wrapper object contains the expected class, and that is basically the same as depending on the sentry module.

### DIFF
--- a/session_redis/http.py
+++ b/session_redis/http.py
@@ -89,6 +89,26 @@ if is_true(os.getenv("ODOO_SESSION_REDIS")):
             host,
             port,
         )
-    http.Application.session_store = session_store
+    target = http.Application
+    if not hasattr(target, "session_store"):
+        # Some other module (at least OCA/server-tools/sentry) has replaced
+        # Application with a different object, hopefully wrapping it instead of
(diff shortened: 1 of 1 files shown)


dnplkndll:
@vrenaville who would review this?

for support with https://github.com/OCA/server-tools/pull/3174
and necessary for volume free deployment

dnplkndll:
this does not seem to actually work. so WIP. it works for some time then started to write to the file store session.

PR #494, opened 2025-05-09
feat: add ssl mode with non url connection + add redis cluster support

vrenaville:


nbessi (approved):

yann-soubeyrand:
@vrenaville @nbessi isn’t it a breaking change to enable SSL by default?

PR #498, opened 2025-09-23
fix: enable module for 19.0

vrenaville:


session_redis/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Sessions in Redis",
     "summary": "Store web sessions in Redis",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
```
leemannd, line 8: This is outdated. The module is migrated and working

logging_json/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "JSON Logging",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
```
leemannd, line 6: AFAIK this is outdated. See
https://github.com/camptocamp/odoo-cloud-platform/pull/502
https://github.com/camptocamp/odoo-cloud-platform/pull/508

But I don't know if it is installed or not on instances

leemannd:
Closing as the PR is outdated. The modules have been migrated.

leemannd (approved):
Work on the modules has also been done into other PRs

session_redis/session.py
```diff
@@ -1,9 +1,10 @@
 # Copyright 2016-2024 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
+import builtins
```
florentx, line 4: Maybe I missed something, but why you need to import `builtins` here?

You could use `list` directly.

If there's a name conflict (I didn't check), in this case, declare `list_type = list` at the start of the file.

session_redis/session.py
```diff
@@ -96,7 +97,7 @@ def save(self, session):
             "utf-8"
         )
         if self.redis.set(key, data):
-            if type(expiration) != int:
+            if not isinstance(expiration, int):
                 expiration = DEFAULT_SESSION_TIMEOUT_ANONYMOUS
             if expiration == 0:
```
florentx: Suggest:
```suggestion
            if not (expiration and isinstance(expiration, int)):
                expiration = DEFAULT_SESSION_TIMEOUT_ANONYMOUS
```

florentx (approved):
LG

session_redis/session.py
leemannd: Added automatically by the precommit

PR #499, opened 2025-09-25
[MIG] session_redis: Migration to 19.0

aisopuro:
* lazy_property is deprecated
* Since https://github.com/odoo/odoo/commit/cea66be976c6d0746a972da8a16b7a239d12a462, Session must be used as a dictionary when setting custom values
* Session default lifetime is now defined in odoo.http, so we can just reuse that
* Handle new session rotation logic (sessions must be findable by the first 42 chars of they key)
* Handle new session key generation for rotation (keys must be longer than 42 chars to handle devices)

### Things I'm still unsure about
* We can import the general session lifetime, but the code also defines `DEFAULT_SESSION_TIMEOUT_ANONYMOUS`, which does not seem to match anything in Odoo (the value is the same as `odoo.http.SESSION_ROTATION_INTERVAL`, but that seems to be a different thing).
* I've changed the store to use the same key generation algorithm used by core Odoo. It seems to me that this is necessary in order to support the new key rotation mechanism for devices, but I'll admit I'm a little unsure. A second pair of eyes would be greatly appreciated.
* I haven't tested this super thoroughly, I'm not sure how.
  * I had to turn off SSL, as I couldn't get it to work in my development environment
  * I haven't actually tested this with any devices, all of this is just based on what the code says and does

### DIFF
--- a/session_redis/__manifest__.py
+++ b/session_redis/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Sessions in Redis",
     "summary": "Store web sessions in Redis",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
     "author": "Camptocamp,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Extra Tools",
--- a/session_redis/http.py
+++ b/session_redis/http.py
@@ -1,12 +1,11 @@
 # Copyright 2016-2024 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
-
+import functools
 import logging
 import os
 
 from odoo import http
 from odoo.tools import config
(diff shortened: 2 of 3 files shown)


aisopuro:
@vrenaville and @nbessi, you reviewed the previous migration of this module to 18. Would it be possible for you to take a look at this 19 migration as well?

cyrilmanuel (approved):

vrenaville:
@aisopuro 
Thanks for your prosition, it seems fine for me. I will ask an other look from our archi @gurneyalex

nbessi (approved):

nbessi:
Hello, 

Looks good to me. I'm not totally sure about the rotate part of the sessions. That looks good on the code, but it will need to be tested live more deeply. Thanks a lot for the work :) For me we can merge

PR #511, opened 2026-02-16
[PERF] fix performance hit

gurneyalex:
On the cloud platform of Camptocamp, a shared redis is used to store the sessions of the different projects -> the number of keys is huge, and using an iterative match kills the performance because of the networking overhead.

We switch to using redis.key(pattern), and since the pattern typically has a leading string which will allow redis to find the correct bucket, the performance should be good.

### DIFF
--- a/session_redis/session.py
+++ b/session_redis/session.py
@@ -176,17 +176,10 @@ class RedisSessionStore(SessionStore):
         identifiers = set(identifiers)
         not_found = set()
         for partial_sid in identifiers:
-            try:
-                next(
-                    self.redis.scan_iter(
-                        match=f"{self.prefix}{partial_sid}*",
-                        count=1,
-                    )
-                )
-            except StopIteration:
-                # No matches found
+            key = f"session::{self.prefix}:{partial_sid}*"
+            match = self.redis.keys(pattern=key)
+            if not match:
                 not_found.add(partial_sid)
-
         return not_found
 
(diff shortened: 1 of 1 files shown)


vrenaville (approved):

fsismondi (approved):

session_redis/session.py
```diff
@@ -176,17 +176,10 @@ def get_missing_session_identifiers(
...
-            try:
-                next(
-                    self.redis.scan_iter(
-                        match=f"{self.prefix}{partial_sid}*",
-                        count=1,
-                    )
-                )
-            except StopIteration:
-                # No matches found
+            key = f"session::{self.prefix}:{partial_sid}*"
```
guewen, line 179: Hi @gurneyalex , wondering why it is not `f"{self.prefix}{partial_sid}*`, as `session:` is already in `self.prefix`. Also the key generation has a quirk to pay attention to: when there is no `ODOO_SESSION_REDIS_PREFIX`, the key is generated as `session:0424149fe72a3e5ae5bd4e3789e80401230ce219` (so no `::`) and when there is one it looks like: `session::foo:0424149fe72a3e5ae5bd4e3789e80401230ce219`.

I'm doing a backport for 18.0 (https://github.com/camptocamp/odoo-cloud-platform/pull/530) and I found that I had to change that, but that's maybe related to 19.0? About the backport, it's still in draft because I will test it live before.

PR #519, opened 2026-03-30
[18.0][REF] cloud_platform,monitoring_prometheus: server_environment should not be a required dependency

yankinmax:
PR is aimed to remove `server_environment` dependency from `cloud_platform` and `monitoring_prometheus` in the scope of this issue:
- [ ] https://github.com/OCA/server-env/issues/258

https://camptocamp.atlassian.net/browse/BSRD-887

### DIFF
--- a/cloud_platform/models/cloud_platform.py
+++ b/cloud_platform/models/cloud_platform.py
@@ -6,7 +6,6 @@ import os
 import re
 
 from odoo import api, models
-from odoo.tools.config import config
 
 from .strtobool import strtobool
 
@@ -22,7 +21,7 @@ class CloudPlatform(models.AbstractModel):
     _description = "cloud.platform"
 
     def _get_running_env(self):
-        environment_name = config["running_env"]
+        environment_name = os.environ.get("RUNNING_ENV")
         if environment_name.startswith("labs"):
             # We allow to have environments such as 'labs-logistics'
             # or 'labs-finance', in order to have the matching ribbon.
--- a/monitoring_prometheus/models/ir_http.py
+++ b/monitoring_prometheus/models/ir_http.py
@@ -17,6 +17,8 @@ class IrHttp(models.AbstractModel):
 
     @classmethod
     def _dispatch(cls, endpoint):
+        # httprequest environment is updated with WSGI environment variables in core
+        # REF: https://github.com/odoo/odoo/blob/18.0/odoo/http.py#L1992
         path_info = request.httprequest.environ.get("PATH_INFO")
 
         if path_info.startswith("/longpolling/"):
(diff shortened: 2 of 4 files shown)


yankinmax:
Hello @ivantodorovich 
I'm not a big expert in these modules.
Following the issue raised and the card description I've opened this PR.
Regarding the `monitoring_statsd` I wonder if we can split it. The way it's implemented required `server_environment` included, otherwise it doesn't make any sense to have it declared.
Other modules in the repo don't have direct/indirect dependency.

Can you pls take a look? 

I think if these two will be validated for the `server_environment` removal we need to decide if we need to forward-port it to v16 and 17?

cloud_platform/models/cloud_platform.py
```diff
@@ -2,74 +2,21 @@
...
-def is_true(strval):
-    return bool(strtobool(strval or "0"))
-
-
 class CloudPlatform(models.AbstractModel):
     _name = "cloud.platform"
     _description = "cloud.platform"
 
-    def _get_running_env(self):
-        environment_name = config["running_env"]
```
ivantodorovich, line 25: Since this is a c2c platform addon, and as such [we always set the `running_env` config through an environment variable](https://github.com/camptocamp/docker-odoo-project/blob/9a4383f3c4d75a9f9473022b13cc7187884650f1/16.0/templates/odoo.cfg.tmpl#L28), and given this is the only method that truly requires `server_environment`, I think we might we able to drop the dependency entirely -- without splitting.

We could just have this line changed, something like this:

```suggestion
        environment_name = os.environ.get("RUNNING_ENV")
```

Cause, even if we drop `server_environment`, we will definitely keep this env variable

ivantodorovich (changes requested):
For `cloud_platform`, see my comment below. I believe we can drop the dependency altogether.

For `monitoring_prometheus`, I don't really understand the split. The extracted things don't really seem to depend on server environment. I think the `server_environment` dependency might have not been real there, and you can remove too

cloud_platform/models/cloud_platform.py
yankinmax: I tend to agree with you.
I've not found any "hard" dependency on `server_environment`, only the need to have an `environment_name` which is for whatever reason implemented different way than any other env in the code.

But, as I don't have any relevant exp with platform modules I wasn't sure.
I'll update this PR, thanks for the review @ivantodorovich

monitoring_prometheus/models/ir_http.py
```diff
@@ -17,6 +17,9 @@ class IrHttp(models.AbstractModel):
 
     @classmethod
     def _dispatch(cls, endpoint):
+        # This doen't require ``server_environment`` module dependency
```
ivantodorovich:
```suggestion
```
no need to mention `server_environment` anymore IMO

but the other comment lines addition for clarification are nice

ivantodorovich (approved):

ivantodorovich (approved):

gurneyalex (approved):

PR #522, opened 2026-03-31
[19.0][REF] cloud_platform,monitoring_prometheus: server_environment should not be a required dependency

yankinmax:
PR is aimed to remove `server_environment` dependency from `cloud_platform` and `monitoring_prometheus` in the scope of this issue:
- [ ] https://github.com/OCA/server-env/issues/258

https://camptocamp.atlassian.net/browse/BSRD-887

Forward-port of:
- [ ] https://github.com/camptocamp/odoo-cloud-platform/pull/519

### DIFF
--- a/cloud_platform/__manifest__.py
+++ b/cloud_platform/__manifest__.py
@@ -9,13 +9,7 @@
     "author": "Camptocamp,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "Extra Tools",
-    "depends": [
-        "session_redis",
-        "monitoring_status",
-        "logging_json",
-        "server_environment",  # OCA/server-tools
-    ],
+    "depends": ["session_redis", "monitoring_status", "logging_json"],
     "website": "https://github.com/camptocamp/odoo-cloud-platform",
-    "data": [],
     "installable": False,
 }
--- a/monitoring_prometheus/__manifest__.py
+++ b/monitoring_prometheus/__manifest__.py
@@ -7,14 +7,9 @@
     "version": "18.0.1.0.0",
     "author": "Camptocamp,Odoo Community Association (OCA)",
     "license": "AGPL-3",
-    "category": "category",
-    "depends": [
-        "base",
-        "web",
-        "server_environment",
-    ],
+    "category": "Extra Tools",
+    "depends": ["base", "web"],
     "website": "https://github.com/camptocamp/odoo-cloud-platform",
-    "data": [],
     "external_dependencies": {
         "python": ["prometheus_client"],
     },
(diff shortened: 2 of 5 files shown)


yankinmax:
This should be checked after the 19.0 branch is activated (modules were `"installable": False`).

monitoring_prometheus/__manifest__.py
```diff
@@ -4,19 +4,14 @@
...
-    ],
+    "category": "Extra Tools",
+    "depends": ["base", "web"],
     "website": "https://github.com/camptocamp/odoo-cloud-platform",
-    "data": [],
     "external_dependencies": {
         "python": ["prometheus_client"],
     },
-    "installable": False,
+    "installable": True,
```
ivantodorovich: do not change that

monitoring_prometheus/__manifest__.py
```diff
@@ -4,19 +4,14 @@
 
 {
     "name": "Monitoring: Prometheus Metrics",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.1",
```
ivantodorovich: do not change it, since the module is not fully migrated and probably won't be

cloud_platform/__manifest__.py
```diff
@@ -5,17 +5,11 @@
...
-        "session_redis",
-        "monitoring_status",
-        "logging_json",
-        "server_environment",  # OCA/server-tools
-    ],
+    "depends": ["session_redis", "monitoring_status", "logging_json"],
     "website": "https://github.com/camptocamp/odoo-cloud-platform",
-    "data": [],
-    "installable": False,
+    "installable": True,
```
ivantodorovich: do not change that

ivantodorovich (changes requested):
Just forward port the commit, but please do not change installable to True, nor update the manifest version.

These modules are not migrated so we don't need to change that here.

cloud_platform/__manifest__.py
```diff
@@ -5,17 +5,11 @@
 {
     "name": "Cloud Platform",
     "summary": "Addons required for the Camptocamp Cloud Platform",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
```
ivantodorovich: don't change the major version -- this means the module is not migrated yet

monitoring_prometheus/__manifest__.py
```diff
@@ -4,17 +4,12 @@
 
 {
     "name": "Monitoring: Prometheus Metrics",
-    "version": "18.0.1.0.0",
+    "version": "19.0.1.0.0",
```
ivantodorovich: don't change the major version -- this means the module is not migrated yet

ivantodorovich (changes requested):

gurneyalex (approved):

ivantodorovich (approved):

PR #523, opened 2026-03-31
[17.0][IMP] attachment_azure: downgrade ResourceExistsError log level

vvrossem:
Using _logger.exception for a ResourceExistsError is misleading: the operation is harmless since blobs are content-addressed (SHA1) and filename is returned correctly regardless. Switch to _logger.info to avoid false-positive ERROR noise.


vvrossem:
superseded by https://github.com/camptocamp/odoo-cloud-platform/pull/524

PR #529, opened 2026-09-03
[ADD] monitoring_prometheus_queue_job

gurneyalex:
This commit features a refactoring of monitoring_prometheus to get a cron gathering some metrics from the database, and then publishing the result on the /metrics endpoint

The new module monitoring_prometheus_queue_job extends the cron to get information about the queue job states.

### DIFF
--- a/monitoring_prometheus/controllers/prometheus_metrics.py
+++ b/monitoring_prometheus/controllers/prometheus_metrics.py
@@ -1,12 +1,54 @@
 # Copyright 2016-2021 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
-from prometheus_client import generate_latest
+import json
+import logging
 
-from odoo.http import Controller, route
+from prometheus_client import Gauge, generate_latest
+
+from odoo.http import Controller, request, route
+
+from ..models.prometheus_gatherer import METRICS_PARAM
+
+_logger = logging.getLogger(__name__)
+
+# Gauges must be instantiated only once per process, registering the same
+# metric name twice raises a "Duplicated timeseries" error.
+_GAUGES = {}
+
+
+def _get_gauge(name, documentation, label_names):
+    gauge = _GAUGES.get(name)
+    if gauge is None:
+        gauge = Gauge(name, documentation, label_names)
+        _GAUGES[name] = gauge
+    return gauge
 
 
 class PrometheusController(Controller):
+    def _publish_gathered_metrics(self):
+        param = request.env["ir.config_parameter"].sudo().get_param(METRICS_PARAM)
+        if not param:
+            return
+
+        cleared = set()
+        for metric in json.loads(param):
+            name = metric["name"]
+            labels = metric.get("labels") or {}
+            gauge = _get_gauge(name, metric.get("documentation", ""), sorted(labels))
+            if name not in cleared:
+                # drop the series that disappeared since the last collection
+                gauge.clear()
+                cleared.add(name)
+            if labels:
+                gauge.labels(**labels).set(metric["value"])
+            else:
+                gauge.set(metric["value"])
+
     @route("/metrics", auth="public")
     def metrics(self):
+        try:
+            self._publish_gathered_metrics()
+        except Exception:
+            _logger.exception("Could not publish the gathered Prometheus metrics")
         return generate_latest()
--- a/monitoring_prometheus/models/prometheus_gatherer.py
+++ b/monitoring_prometheus/models/prometheus_gatherer.py
@@ -1,0 +1,48 @@
+# Copyright 2016-2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+import json
+
+from odoo import api, models
+
+METRICS_PARAM = "monitoring_prometheus.metrics"
+
+
+class PrometheusGatherer(models.AbstractModel):
+    """Collect application metrics to be exposed on the /metrics endpoint.
+
+    Metrics are gathered by a cron and stored as JSON in an
+    ``ir.config_parameter``, because crons and HTTP workers run in
+    separate processes and cannot share the Prometheus registry.
+    """
+
+    _name = "prometheus.gatherer"
+    _description = "Prometheus Metrics Gatherer"
+
+    @api.model
+    def _gather_metrics(self):
+        """Return the metrics as a list of dicts.
+
+        Each entry is of the form::
+
+            {
+                "name": "odoo_some_metric",
+                "documentation": "What this metric measures",
+                "labels": {"label_name": "label_value"},
+                "value": 42,
+            }
+
+        ``labels`` may be omitted for an unlabelled metric. All the entries
+        sharing a ``name`` must declare the same label names.
+
+        Modules extending this method must return the result of ``super()``
+        extended with their own entries.
+        """
+        return []
+
+    @api.model
+    def _cron_gather_metrics(self):
+        self.env["ir.config_parameter"].sudo().set_param(
+            METRICS_PARAM, json.dumps(self._gather_metrics())
+        )
+        return True
--- a/monitoring_prometheus_queue_job/__manifest__.py
+++ b/monitoring_prometheus_queue_job/__manifest__.py
@@ -3,14 +3,15 @@
 
 
 {
-    "name": "Monitoring: Prometheus Metrics",
-    "version": "14.0.1.1.0",
+    "name": "Monitoring: Prometheus Metrics for Queue Jobs",
+    "version": "14.0.1.0.0",
     "author": "Camptocamp,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "category",
-    "depends": ["base", "web", "server_environment"],
+    "depends": ["monitoring_prometheus", "queue_job"],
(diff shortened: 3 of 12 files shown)


monitoring_prometheus_queue_job/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2016-2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+
+{
+    "name": "Monitoring: Prometheus Metrics for Queue Jobs",
+    "version": "14.0.1.0.0",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "category",
```
twalter-c2c, line 10: Do we want a more accurate category name? 😉

monitoring_prometheus_queue_job/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2016-2021 Camptocamp SA
```
twalter-c2c, line 1:
```suggestion
# Copyright 2026 Camptocamp SA
```

monitoring_prometheus_queue_job/models/prometheus_gatherer.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2016-2021 Camptocamp SA
```
twalter-c2c, line 1:
```suggestion
# Copyright 2026 Camptocamp SA
```

monitoring_prometheus/models/prometheus_gatherer.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2016-2021 Camptocamp SA
```
twalter-c2c, line 1:
```suggestion
# Copyright 2026 Camptocamp SA
```

twalter-c2c (approved):
Looks good, AFAICT. Pre-commit needs to be satisfied. I dropped a few tiny comments, but they are not relevant to the code logic. Pre-approving.

monitoring_prometheus/models/prometheus_gatherer.py
```diff
@@ -0,0 +1,48 @@
...
+from odoo import api, models
+
+METRICS_PARAM = "monitoring_prometheus.metrics"
+
+
+class PrometheusGatherer(models.AbstractModel):
+    """Collect application metrics to be exposed on the /metrics endpoint.
+
+    Metrics are gathered by a cron and stored as JSON in an
+    ``ir.config_parameter``, because crons and HTTP workers run in
```
florentx, line 15: It is interesting, but there's no better choice to store metrics, than `ir.config_parameter`?

I was just curious if it is a pattern used elsewhere.

florentx (approved):

SilvioC2C (approved):
Overall LGTM; wondering if there's a better alternative to storing log values in a sys param (especially cause they can be edited manually from the UI), but definitely not a blocker

monitoring_prometheus/models/prometheus_gatherer.py
gurneyalex: We can materialize a model for this. The main thing is that the value is gathered in one worker (the cron) and served over http, so we need to store it somewhere, we cannot keep it in memory (which is what is done for the http metrics in the module (which are useless btw, but this is a different issue). I also wanted to avoid having the /metrics endpoint make potentially complex SQL queries, to keep it fast. 

I chose ir.config_parameter because it is guaranteed to be present and kind of "made sense" . What I did not check is if there is a cache invalidation involved when we update it (we certainly don't want that) -> will double check.

SilvioC2C: @gurneyalex unfortunately, yes: https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_config_parameter.py#L100-L117

florentx: Possible alternatives could be:
- `ir.logging` which is not used much
- an attachment on `ir.cron` record with id `"monitoring_prometheus.cron_prometheus_gather_metrics"`

It could be a singleton record with an "external ID" to retrieve it with `env.ref("monitoring_prometheus.metrics")` for example and you're done.

monitoring_prometheus/models/prometheus_gatherer.py
```diff
@@ -0,0 +1,48 @@
...
+        ``labels`` may be omitted for an unlabelled metric. All the entries
+        sharing a ``name`` must declare the same label names.
+
+        Modules extending this method must return the result of ``super()``
+        extended with their own entries.
+        """
+        return []
+
+    @api.model
+    def _cron_gather_metrics(self):
```
divad1196, line 44: ## Why the cron

I don't really see the benefit of using a cron here.
If prometheus pull faster than the cron, it will store the same value over and over. 

If there is a reason for that, maybe we should use a collector in push mode.
For example [Prometheus Pushgateway](https://prometheus.io/docs/practices/pushing/) or [OpenTelemetry collector](https://opentelemetry.io/docs/collector/):
- The cron push data directly to it
- Prometheus collect from the collector

Prometheus itself accepts push, but it's usually preferable to keep it in pull mode.
This would at least offload Odoo and also avoid the `ir.config_parameter` trick.

The `/metrics` route only makes sense if we serve fresh data IMO.

## Naming
It's a bit confusing here:
We also have `_gather_queue_job_metrics`, similar name, that actually extend `_gather_metrics`

This method should maybe be called `_cron_persist_metrics`

monitoring_prometheus/controllers/prometheus_metrics.py
```diff
@@ -1,12 +1,54 @@
...
+from ..models.prometheus_gatherer import METRICS_PARAM
+
+_logger = logging.getLogger(__name__)
+
+# Gauges must be instantiated only once per process, registering the same
+# metric name twice raises a "Duplicated timeseries" error.
+_GAUGES = {}
+
+
+def _get_gauge(name, documentation, label_names):
```
divad1196, line 20: Could mention that it's a multiton pattern.
This helps scanning the intent.

## Alternative approach:
```python
registry = CollectorRegistry()
for m in ... : 
     Gauge(m["name"], ... , registry=reg)
```

The key must be unique per registry, and when not specified, `Gauge` uses the global default registry.
To use it, I would group the data by metric name.

### Benefit
Avoiding race condition and side effects

This is technically not and issue in worker mode as it currently process 1 request at the time, but this is an issue in thread mode. And this will be an issue with the hybrid mode.

monitoring_prometheus/controllers/prometheus_metrics.py
```diff
@@ -1,12 +1,54 @@
...
+    gauge = _GAUGES.get(name)
+    if gauge is None:
+        gauge = Gauge(name, documentation, label_names)
+        _GAUGES[name] = gauge
+    return gauge
 
 
 class PrometheusController(Controller):
+    def _publish_gathered_metrics(self):
+        param = request.env["ir.config_parameter"].sudo().get_param(METRICS_PARAM)
```
divad1196, line 30: `param` isn't suitbable for a name

```python
_metrics_dump = ...
if not _metricis_dump: 
    return
metrics = json.loads(_metrics_dump)
```

monitoring_prometheus/controllers/prometheus_metrics.py
```diff
@@ -1,12 +1,54 @@
...
+            return
+
+        cleared = set()
+        for metric in json.loads(param):
+            name = metric["name"]
+            labels = metric.get("labels") or {}
+            gauge = _get_gauge(name, metric.get("documentation", ""), sorted(labels))
+            if name not in cleared:
+                # drop the series that disappeared since the last collection
+                gauge.clear()
```
divad1196, line 41: If a metric is missing, it won't be cleared.
This is an issue in a multi-database server:
Metrics available on database A but not on B will still appear on B and not be cleared up.

We should have the clearup at the start, this would also remove the need for the set:
```python
for g in _GAUGES.values():
    g.clear()
```

Using a new `CollectorRegistry` per request solves the issue

divad1196 (commented):
TL;DR:
- I don't see the benefit of the cron over doing the request on-the-fly when `/metrics` is called.  We can simply configure prometheus to pull only every 15min
- If there is a reason for it, I would prefer to use a prometheus collector and push to it from the cron instead.

sbidoul:
@gurneyalex  OCA/queue#976 may be of interest to you

monitoring_prometheus/models/prometheus_gatherer.py
gurneyalex: I added a prometheus.metric model

monitoring_prometheus/models/prometheus_gatherer.py
gurneyalex: Having a cron and storing the values means that they are collected once and for all, because these values are shared by all the workers which are part of the instance. I was concerned about hitting the database too hard with 15 pods being queried at the same time by prometheus on some instances. Maybe it is not an issue. 

refactored the code for the naming.

monitoring_prometheus/controllers/prometheus_metrics.py
gurneyalex: I'll check this later, unless you beat me to it and propose a PR (I will be away next week, I'm not sure of the importance of this for @florentx 's project

PR #530, opened 2026-09-07
[18.0][FIX] session_redis: backport device session methods

guewen:
Backport from https://github.com/camptocamp/odoo-cloud-platform/pull/499
Partial cherry-pick of commit https://github.com/camptocamp/odoo-cloud-platform/commit/750bc4dcf876cd4b35fd9925b605a87a7a42d49c

The methods have been added on the SessionStore in odoo 18.0 through
https://github.com/odoo/odoo/commit/80224f38f0869dc540bb022fda41e0e6209a412b

Also applied the performance fix done in https://github.com/qoqa/odoo-cloud-platform/commit/911fa5535aafd09fc0fde6935a62449be58da97f

### DIFF
--- a/session_redis/session.py
+++ b/session_redis/session.py
@@ -1,10 +1,12 @@
 # Copyright 2016-2024 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
+import builtins
 import json
 import logging
+from typing import TypeAlias
 
-from odoo.service import security
+import odoo
 from odoo.tools._vendor.sessions import SessionStore
 
 from . import json_encoding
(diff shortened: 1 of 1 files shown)


session_redis/session.py
```diff
@@ -120,3 +126,47 @@ def vacuum(self, *args, **kwargs):
...
+
+        While this method is not part of the generic SessionStore API, it is
+        defined on the file session store, and is used by Odoo's devices to
+        figure out what needs to be revoked
+        (see odoo.addons.base.models.res_device.ResDeviceLog.__update_revoked).
+        """
+        identifiers = set(identifiers)
+        not_found = set()
+        for partial_sid in identifiers:
+            key = f"{self.prefix}{partial_sid}*"
```
guewen, line 145: In branch 19.0, the key is set as `f"session::{self.prefix}:{partial_sid}*"`, which does not look correct to me? because `self.prefix` already contains the `session:` and final `:` parts. A key looks like "session:0424149fe72a3e5ae5bd4e3789e80401230ce219" or "session::foo:0424149fe72a3e5ae5bd4e3789e80401230ce219".

session_redis/session.py
```diff
@@ -120,3 +126,47 @@ def vacuum(self, *args, **kwargs):
...
+        While this method is not part of the generic SessionStore API, it is
+        defined on the file session store, and is used by devices when revoking
+        device sessions (see odoo.addons.base.models.res_device.ResDevice._revoke).
+        """
+        patterns_to_unlink = []
+        for identifier in identifiers:
+            # Avoid removing a session if it does not match an identifier.
+            # See this same comment in
+            # odoo.http.FileSessionStore.delete_from_identifiers.
+            if not _sha1_re.match(identifier):
```
guewen, line 164: To minimize the change, instead of changing the way of generation session sid as done in 19.0, I changed the check to use the pattern that matches the sid generated by `SessionStore` (40 chars).

